PR #141, opened 2023-06-09
Add cond operation for conditional selection

eb8680:
Extracted from #135. ~~Blocked by #140.~~

PR #142, opened 2023-06-09
Add preempt operation

eb8680:
Extracted from #135.
~~Blocked by #141.~~

Tasks:
- [x] Implement `preempt`
- [x] Implement `Preemptions`
- [x] Add documentation
- [x] Add unit tests


rfl-urbaniak (approved):
Also tested on two models (bayesian stones with a categorical outcome variable, and a drug-bp-exercise model with continuous variables only), obtaining the intended results. I can add those as test cases if needed. Some time later on I'd consider including an example in the documentation.

Issue #165, opened 2023-07-03
Improving compatibility with Pyro broadcasting semantics


eb8680:
This issue tracks work toward making `causal_pyro` work correctly and seamlessly with the broadcasting-based implementations of various modeling and inference computations in Pyro.

Points in `causal_pyro` that need new test cases and bugfixes:
- Indexing operations in `causal_pyro.indexed.ops`:
  - [x] `indices_of` - #168 
  - [x] `scatter` - #168 
  - [x] `gather` - #168 
  - [x] `cond` - #141
- Older unit tests in `causal_pyro.counterfactual` that do not use the `indexed` API

Potential points of failure upstream in Pyro that should be fixed by improvements to `indexed` but should still be covered in `causal_pyro` unit tests (it should be straightforward to adapt test cases from Pyro's `test/infer/test_valid_models.py` and `test/infer/test_enum.py`):
- Models and guides:
  - [ ] Nested plate configurations
  - [ ] `max_plate_nesting > 5` (see #46)
  - [ ] `pyro.poutine.mask`
  - [ ] `pyro.poutine.scale`
- Plates used for automatic vectorization in inference:
  - [x] `pyro.infer.Predictive(..., parallel=True)` - #176 
  - [x] `pyro.infer.Trace*_ELBO(..., vectorize_particles=True)` - #176 
  - [x] `pyro.infer.importance.vectorized_importance_weights`
- Enumeration-based ELBO estimators:
  - [x] `pyro.infer.TraceEnum_ELBO` (model-side enumeration) - #176  
  - [x] `pyro.infer.TraceEnum_ELBO` (guide-side enumeration) - #176 
  - [x] `pyro.infer.TraceTMC_ELBO` - #176 
- Other enumeration-based inference:
  - [x] `pyro.infer.TraceEnum_ELBO.compute_marginals` - #176 
  - [x] `pyro.infer.infer_discrete` - #176 
  - [x] `pyro.infer.HMC` - #176 
- Stretch goals
  - `pyro.contrib.funsor`
  - `torch.vmap`

PR #176, opened 2023-07-05
Refactor counterfactuals to use observe and condition

eb8680:
Extracted from #173 
~~Blocked by #175~~
Addresses #165 
~~Blocked by #205~~

This PR includes the following high-level changes:
- Complete overhaul of the current implementation of `causal_pyro.counterfactual.handlers.ambiguity` to use `observe` from #175 instead of working with `pyro.sample` directly
- Use of `condition` from #175 in place of `pyro.condition` throughout the current unit tests
- Tests exercising a number of enumeration cases from #165 
- Notebooks have been re-run to reproduce results under the other changes here (#178 )


eb8680:
The diffs for the notebooks in this PR are large because of git, but the only changes are:
1. Importing `condition` from `chirho.observational.handlers`
2. Replacing all uses of `pyro.condition` with `condition`
3. Re-executing the notebook from scratch to verify that the results are unaffected

azane (commented):
I just looked through the notebooks (and tested the merge locally) and some of the formatting and link fixes to notebooks get reverted by this PR. Am investigating now. I'm going to see if I can cherry-pick the changes back in here, but also may just copy/paste things as it's not a huge amount being reverted.

SamWitty (approved):
This looks great. The change actually makes the multi-world semantics pop out of the implementation much more clearly.

PR #199, opened 2023-07-10
Rename observational module to constrained

eb8680:
~~Blocked by #176~~

This PR renames `chirho.observational` to `chirho.constrained`, which is a little more standard.


eb8680:
Closing for now - unclear if this is actually an improvement

PR #218, opened 2023-07-12
Cleanup of Performance Improvements for ODE Conditioning

azane:
Blocking #219 

Does not implement handler composition tests. That's captured in issue #149.

Issue #149, opened 2023-06-12
Unit tests for composing intervention, observation and counterfactual handlers


agrawalraj:
- Write tests that compose `PointObservation` and `PointIntervention` in #137  with other Pyro handlers
    - E.g. wrap a model in a plate and see if it works
    - E.g. enumeration

- [x] #219
- [x] #214 (This includes intervention/counterfactual composition tests.
- [x] Test of composition of multi-world counterfactual.

PR #219, opened 2023-07-12
Test Composition of Dynamic Counterfactual and Observation

azane:
~~Holding off on this b/c the dynamic demo doesn't actually compose these things, it just performs posterior inference and then projects the posterior predictive forward through counterfactuals in a separate handled execution. I.e. counterfactuals and observation handlers don't appear in the same stack right now.~~

Notebook has been modified to include counterfactual queries requiring inference on post-intervention noise, so these do appear in the same call stack now.

~~Blocked by #218.~~

Addresses testing composing counterfactuals and observation handlers. See issue #149


tests/dynamical/dynamical_fixtures.py
```diff
@@ -6,48 +6,43 @@
 from chirho.dynamical.ops import State, Trajectory
 
 
-class SimpleSIRDynamics(ODEDynamics):
-    @pyro.nn.PyroParam(constraint=constraints.positive)
-    def beta(self):
-        return torch.tensor(0.5)
+class UnifiedFixtureDynamics(ODEDynamics):
```
azane, line 9: Modified so there's only one place in which shapes need to be manually managed.

agrawalraj (approved):

PR #220, opened 2023-07-13
Fix PyroModule state persistence in simulate

eb8680:
This PR ensures that the persistence of `PyroSample` state in `PyroModule` is guaranteed across an entire effectful `simulate` call, even in the presence of a `SimulatorEventLoop`.

### DIFF
--- a/chirho/dynamical/handlers.py
+++ b/chirho/dynamical/handlers.py
@@ -37,7 +37,7 @@ class ODEDynamics(pyro.nn.PyroModule):
 
     # noinspection PyMethodParameters
     def _deriv(
-        dynamics: "ODEDynamics",
+        self: "ODEDynamics",
         var_order: Tuple[str, ...],
         time: torch.Tensor,
         state: Tuple[torch.Tensor, ...],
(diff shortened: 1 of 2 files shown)


chirho/dynamical/handlers.py
```diff
@@ -46,9 +46,31 @@ def _deriv(
...
         for var, value in zip(var_order, state):
             setattr(env, var, value)
-        dynamics.diff(ddt, env)
+        assert "t" not in env.keys, "variable name t is reserved for time"
+        env.t = time
+        self.diff(ddt, env)
         return tuple(getattr(ddt, var, torch.tensor(0.0)) for var in var_order)
 
+    @pyro.nn.pyro_method
+    @pyro.poutine.runtime.effectful(type="simulate")
```
eb8680, line 53: The key change in this PR is the presence of `pyro.nn.method` outside the `effectful` decorator on this method of `ODEDynamics`. This ensures that `PyroSample` statements have a single value throughout the entire `simulate` call.

tests/dynamical/test_static_observation.py
```diff
@@ -246,3 +246,32 @@ def conditioned_sir(times_data):
     pyro.clear_param_store()
     for step in range(n_steps):
         svi.step((times, data))
+
+
+@pytest.mark.parametrize("use_event_loop", [True, False])
+def test_simulate_persistent_pyrosample(use_event_loop):
```
eb8680, line 252: This test fails on `staging-dynamic` and passes with the fix in this PR.

azane (approved):

PR #223, opened 2023-07-14
Added counterfactual example to the demo

agrawalraj:
~~Blocked by #219.~~

We added the section "Modeling a superspreader event using counterfactual inference" to the notebook which performs counterfactual inference in a dynamical system governed by an SIR model.

Pair coded with @azane @eb8680


azane:
Tests will pass once #219 is merged (and then staging-dynamic is merged into this).

eb8680 (approved):
The final story looks great. We can go back and add explanatory text later once some of the many design issues we uncovered with the dynamical systems interfaces are addressed.

PR #224, opened 2023-07-14
Added test to check output of estimating counterfactual with AutoSoftConditioning

agrawalraj:
This tests considers the following model:
```
def model():
        z = pyro.sample("z", dist.Normal(0, 1), obs=torch.tensor(0.1))
        u_x = pyro.sample("u_x", dist.Normal(0, 1))
        x = pyro.deterministic("x", z + u_x, event_dim=0)
        u_y = pyro.sample("u_y", dist.Normal(0, 1))
        y = pyro.deterministic("y", x + z + u_y, event_dim=0)
        return dict(x=x, y=y, z=z)
```
It conditions on the values `z=0.1`, `x=0`, and `y=1`. Then, 

$$ p(u_x \mid z=.1, x=0, y=1) = I(u_{x}=(x - z)) =  I(u_{x} = (0 - .1)) = I(u_{x} = -0.1) $$

and

$$p(u_y \mid z=.1, x=0, y=1) = I(u_{y}=(y - x - z)) =  I(u_{y} = (1 - 0 - .1)) = I(u_{y} = 0.9).$$

This test compares the output of SVI using `AutoSoftConditioning` to handle the deterministic sites resulting from reparameterizing the model to share noise variables in `TwinWorldCounterfactual`. Specifically, this test requires that the posterior mean is "close" to the ground truth values above, where "close" is determined by the scale used in `AutoSoftConditioning`.

### DIFF
--- a/tests/counterfactual/test_counterfactual_handler.py
+++ b/tests/counterfactual/test_counterfactual_handler.py
@@ -6,6 +6,7 @@ import pyro.distributions as dist
 import pyro.infer
 import pytest
 import torch
+from pyro.infer.autoguide import AutoMultivariateNormal
 
 import chirho.interventional.handlers  # noqa: F401
 from chirho.counterfactual.handlers import (  # TwinWorldCounterfactual,
@@ -19,7 +20,10 @@ from chirho.indexed.ops import IndexSet, gather, indices_of, union
 from chirho.interventional.handlers import do
 from chirho.interventional.ops import intervene
 from chirho.observational.handlers import condition
+from chirho.observational.handlers.soft_conditioning import AutoSoftConditioning
 from chirho.observational.ops import observe
+
+pyro.settings.set(module_local_params=True)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


azane:
@agrawalraj do you think this will have like a 3% failure rate or 0.1% failure rate or etc. etc.

SamWitty:
Generally I'm nervous about adding full inference runs to test suites because:

1. It can take a long time and bog down CI builds, etc.
2. As @azane is alluding to, they can fail with moderately high probability

@agrawalraj , is there a way to test the functionality you want without running SVI?

SamWitty:
@eb8680 , what is your general thinking about these kinds of integration tests including inference?

azane:
@SamWitty re: your first point 

> It can take a long time and bog down CI builds

This one only takes < 4 seconds on Raj's machine locally.

agrawalraj:
> Generally I'm nervous about adding full inference runs to test suites because:
> 
> 1. It can take a long time and bog down CI builds, etc.
> 2. As @azane is alluding to, they can fail with moderately high probability
> 
> @agrawalraj , is there a way to test the functionality you want without running SVI?

I would like to test the functionality of actually recovering the noise variables correctly. In the example, the noise distributions are just point masses. I think testing recovery would require some inference procedure...

eb8680 (approved):
This seems fine, I think it became clear at the hackathon that having a few stronger tests like this for counterfactuals was extremely helpful.

PR #226, opened 2023-07-24
both factual outcomes -> both outcomes

djinnome:
The factual outcome is what actually happened, not the counterfactual outcome.


eb8680 (approved):
Thanks for the fix!

PR #227, opened 2023-07-27
Fix Gradient Propagation Through Dynamic Interventions

azane:
This was not working b/c timespans were being spliced with `torch.tensor` (which apparently clears out gradients) instead of `torch.cat` which preserves them. Two tests added — one of the behind-the-scenes torchdiffeq functionality, and another doing the exact same thing but using `DynamicIntervention`.

Note that gradients do not propagate through the `PointIntervention` `time` argument. I added a test for this but marked it as skipped for now. Added issue #228 to track this.

### DIFF
--- a/tests/dynamical/test_dynamic_interventions.py
+++ b/tests/dynamical/test_dynamic_interventions.py
@@ -2,19 +2,22 @@ import logging
 
 import pytest
 import torch
+from torch import tensor as tt
 
 from chirho.counterfactual.handlers import (
     MultiWorldCounterfactual,
     TwinWorldCounterfactual,
 )
-from chirho.dynamical.handlers import DynamicIntervention, SimulatorEventLoop, simulate, PointIntervention
+from chirho.dynamical.handlers import (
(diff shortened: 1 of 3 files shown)


tests/dynamical/test_dynamic_interventions.py
```diff
@@ -317,3 +323,151 @@ def test_split_twinworld_dynamic_matches_output(
...
+
+    # Z begins accruing dz=1 at t=param, so dzdparam should be -1.0.
+    (dzdparam,) = torch.autograd.grad(
+        outputs=(traj.z[-1],), inputs=(param,), create_graph=True
+    )
+    assert torch.isclose(dzdparam, tt(-1.0), atol=1e-5)
+
+
+@pytest.mark.skip(reason="TODO")
+def test_grad_of_point_intervention_params():
```
azane, line 368: This is the skipped point intervention test. See issue #228.

agrawalraj (approved):
Great catch!

Issue #26, opened 2022-10-18
Add support for additional intervention types


eb8680:
In classical causal inference theory, interventions are defined as replacements of constant values with constant values, but many advanced causal queries can be understood as arising from a richer notion of intervention, e.g. replacing one mechanism with another or replacing a value with a distribution.

Fortunately, most of these interventions are fully compatible in principle with the broadcasting-based counterfactual semantics of this library, and queries that make use of richer interventions can therefore be answered with Pyro's existing inference tools.

All that is necessary to implement this support is to extend the definition of the existing `intervene` primitive and the `MultiWorldCounterfactual` effect handler to new input and output types.

Intervention types:
- [x] Atomic values: `torch.Tensor` or `numbers.Number`
- [x] Dependent interventions: `typing.Callable[[T], Intervention[T]]`
- [x] #113 Interventions on mechanisms: `typing.Callable[[T], T]` or `pyro.nn.PyroModule`
- [x] #121 Multiple simultaneous interventions: `typing.Tuple[T, ...]`
- [ ] Stochastic interventions: `pyro.distributions.TorchDistribution` or `pyro.nn.PyroSample`
- [ ] Free parameter interventions: `pyro.nn.PyroParam`

PR #239, opened 2023-08-31
Add a BiasedPreemptions handler

eb8680:
Part of modularizing #236 

This PR adds a new effect handler `BiasedPreemptions` for inserting `chirho.counterfactual.ops.preempt` calls after `pyro.sample` calls. Unlike `Preemptions`, this handler flips a possibly unfair coin to decide whether or not an intervention takes place.

There are also some minor changes to the organization of default `preempt` behavior across `chirho.counterfactual.handlers.MultiWorldCounterfactual` and `chirho.counterfactual.handlers.counterfactual.Preemptions`.

Tested:
- Added a test case to existing `Preemptions` unit test

### DIFF
--- a/chirho/counterfactual/handlers/counterfactual.py
+++ b/chirho/counterfactual/handlers/counterfactual.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Generic, Mapping, Tuple, TypeVar
+from typing import Any, Dict, Generic, Mapping, TypeVar
 
 import pyro
 import torch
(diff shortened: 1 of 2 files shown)


chirho/counterfactual/handlers/counterfactual.py
```diff
@@ -100,23 +105,102 @@ class Preemptions(Generic[T], pyro.poutine.messenger.Messenger):
...
 
-    def __init__(self, actions: Mapping[str, Intervention[T]]):
+    def __init__(
+        self, actions: Mapping[str, Intervention[T]], *, prefix: str = "__split_"
+    ):
         self.actions = actions
+        self.prefix = prefix
         super().__init__()
 
-    def _pyro_post_sample(self, msg: Dict[str, Any]) -> None:
```
eb8680, line 111: There's a bit of mismatch in the way `split` and `preempt` unpack their arguments versus `intervene` and the structure of the complex `Intervention` type in `chirho.interventional.ops`. I removed the annotation here to prevent `mypy` from causing CI failures over this minor issue, which can be dealt with directly in a separate PR.

eb8680: This issue is related to and perhaps subsumed by #26

PR #241, opened 2023-08-31
Operations and Handlers for Dynamical Systems

SamWitty:
Initial PR for operations and handlers supporting causal/probabilistic inference in continuous time dynamical systems.

This PR includes the following new user-facing classes:

`State`: Representation of the collection of variables at a given moment in time. E.g. "Susceptible", "Infected", etc.
`Trajectory`: Representation of a collection of states across multiple moments in time.
`Dynamics`: Abstract representation of the continuous time dynamical systems model. `Dynamics` subclasses must implement two methods, `diff` representing the discrete or continuous difference in state over time, and `observation` representing the process by which data is generated noisily from dynamical system trajectories.

This PR introduces a new dynamical system-specific effectful operation:

`simulate`: A function that takes as input an instance of a `Dynamics` object, a collection of simulation arguments, and optionally a `solver` (currently just `TorchDiffEq`) if there is no enclosing `solver` effect handler. `simulate` returns the final state of the simulation, but can be used in the context of a `DynamicTrace` handler to extract entire trajectories instead.

Also, this PR provides a type-directed dispatch for new classes on the following ChiRho operations: `intervene`, `indices_of`, and `gather`.

This PR also includes a collection of handlers for orchestrating numerical simulation, applying interventions, and conditioning on observations at moments in continuous time. Specifically, we introduce the following 

Effect Handlers:

`PointInterruption`: Generic handler class for splitting simulations at a moment in time.
`PointIntervention`: Handler for applying an intervention at a given moment in time. Currently, this is implemented by replacing the state at that time point, which contains all attributes defined in the model's diff method.
`PointObservation`: Handler for observing noisy measurements of state at a moment in time. Similar to pyro.condition, which handler fixes the value of the observed random variable, and adds to a global log likelihood store for subsequent inference.
`DynamicIntervention`: Handler for applying state-dependent interventions
`SimulatorEventLoop`: Effect handler for aggregating and orchestrating simulations interspersed by PointInterruptions.
`DynamicTrace`: Effect handler for aggregating trajectories from a (sequence of) `simulate` calls.
`NonInterruptingPointObservationArray`: Implementation of conditioning that subclasses `DynamicTrace`, avoiding quadratic overhead from interrupting `simulate` calls with multiple handler instances.
`Solver`: Effect handler for specifying the particular dynamical system solver for any encapsulated `simulate` calls. Required when `simulate`'s `solver` argument is left unassigned.
`TorchDiffEq`: Subclass of `Solver` for solving ODEs with the torchdiffeq package. https://github.com/rtqichen/torchdiffeq

Notebook:
`dynamical_intro.ipynb`: notebook documenting the basic functionality of defining and learning dynamical systems using the newly defined classes and handlers described above


SamWitty:
Blocked by #288 and #289

PR #242, opened 2023-08-31
fixed test to reflect PyroModule changes

SamWitty:
This small PR modifies some dynamical systems tests to use the more modern `PyroModule` interface for specifying probabilistic/causal models in Pyro/ChiRho. This avoids bugs introduced by using the preferable `pyro.settings.set(module_local_params=True)` when using SVI.


eb8680 (approved):
LGTM

PR #250, opened 2023-09-04
Replace Preemptions handler implementation

eb8680:
This PR removes the original `Preemptions` handler added in #143 and replaces it with the renamed implementation of `BiasedPreemptions` from #239, which is equivalent when the bias parameter is set to its default value of 0.

`Preemptions` is currently not being used on `master` or in any of the latest causality code. It's also not clear to me when and how we would have used `Preemptions` in a way distinct from `BiasedPreemptions`, or that the approach in the original version of `Preemptions` of outsourcing default behavior for `preempt` was sensible to begin with.

Tested:
- Refactoring covered by existing unit tests

### DIFF
--- a/tests/counterfactual/test_counterfactual_handler.py
+++ b/tests/counterfactual/test_counterfactual_handler.py
@@ -15,7 +15,7 @@ from chirho.counterfactual.handlers import (  # TwinWorldCounterfactual,
     SingleWorldFactual,
     TwinWorldCounterfactual,
 )
-from chirho.counterfactual.handlers.counterfactual import BiasedPreemptions, Preemptions
+from chirho.counterfactual.handlers.counterfactual import Preemptions
 from chirho.counterfactual.handlers.selection import SelectFactual
 from chirho.counterfactual.ops import preempt, split
 from chirho.indexed.ops import IndexSet, gather, indices_of, union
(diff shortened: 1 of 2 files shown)


tests/counterfactual/test_counterfactual_handler.py
```diff
@@ -693,12 +692,7 @@ def model():
...
 
-    if use_biased_preemption:
-        preemption_handler = BiasedPreemptions(
-            actions=preemptions, bias=0.1, prefix="__split_"
-        )
-    else:
-        preemption_handler = Preemptions(actions=preemptions)
+    preemption_handler = Preemptions(actions=preemptions, bias=0.1, prefix="__split_")
 
     with MultiWorldCounterfactual(cf_dim), preemption_handler:
```
rfl-urbaniak, line 697: I'd also consider testing whether the intended preemptions have been implemented and whether their frequency is more or less as intended, maybe along the lines of:

        intervened_y = gather(tr.nodes["y"]["value"], IndexSet(x={1}))
        preempted_y_freq = torch.tensor((intervened_y == 1.0).sum().item() / 1000)
        assert torch.isclose(preempted_y_freq, torch.tensor(0.6), atol=0.09)

rfl-urbaniak (changes requested):

rfl-urbaniak (approved):

PR #252, opened 2023-09-05
Decouple ODEDynamics class from torchdiffeq solver

SamWitty:
This refactoring PR decouples the definition of the `ODEDynamics` with the specific solver used to implement `simulate` and adds additional structure to the module to be more consistent with existing `chirho` modules. In addition, this refactor removes some indirection in dispatch mechanisms that impedes readability.

Specifically, this PR makes the following changes:

1. Migrate all non-effectful classes/objects into a corresponding `internals.py` files. This includes anything that isn't directly defining an operation, a dispatch on a default implementation of an operation, or a handler. E.g. `Dynamics`, `ODEDynamics`, `State`, etc.
3. Explicitly distinguish between effectful operations (e.g. `simulate`) which are not dispatched on type and the redirected default implementation (e.g. `_simulate`) which are dispatched on the type of the first argument. This avoids messy interactions between `functools.singldispatch` and `pyro.effectful` decorators and also avoids confusion when defining functions in `ops.py` files that don't have `pyro.effectful` decorators.
4. Add additional file structure for denoting a hierarchy of dynamical systems formalisms. `ODE` inherits from generic `dynamics`, and includes a collection of specific solvers in the `ODE.backends` module.
5. Remove support for calling `simulate(...)` out of context of a particular simulation backend class that inherits from the generic `SimulationEventLoop`. Added a test to `test_noop_interruption.py` to confirm this.
6. Refactor existing backend using `torchdiffeq` into an explicit `TorchDiffEq` handler.


eb8680:
In `chirho/dynamical/handlers.py`, `_InterventionMixin` and other handler types should be parametrized by a type variable, rather than specific to `torch.Tensor`.

chirho/dynamical/ops.py
```diff
@@ -235,12 +122,25 @@ def trajectory_concatenate(*trajectories: Trajectory[T], **kwargs) -> Trajectory
...
+def _simulate(
     dynamics: Dynamics[S, T], initial_state: State[T], timespan, **kwargs
 ) -> Trajectory[T]:
     """
     Simulate a dynamical system.
     """
     raise NotImplementedError(f"simulate not implemented for type {type(dynamics)}")
+
+
+_simulate.register(ODEDynamics, _ode_simulate)  # type: ignore
```
eb8680, line 146: To make this slightly more ergonomic, you could set
```
simulate.register = _simulate.register
```

eb8680:
@SamWitty can this be closed in favor of #290?

SamWitty:
Yes.

Issue #243, opened 2023-09-01
Docs build is failing upon deployment to GH pages


eb8680:
This first happened after #238 was merged, but I haven't had any issues building docs locally and the CI docs build worked fine in #238. It's probably something to do with the versions of the requirements in `docs/requirements.txt`.

The previously deployed version of the docs is still up, so we should resolve this soon but it doesn't have to happen immediately.

Log here: https://github.com/BasisResearch/chirho/actions/runs/6049890251/job/16425678930

SamWitty:
I'll hunt down the culprit dependencies.

eb8680:
Two things to try are building from the commit prior to #238 to see if it still fails and pinning the PyTorch and Pyro versions.

SamWitty:
Tried checking out the commit prior to #238 and pinning PyTorch and Pyro versions and still no success. I also went with the nuclear option and pinned all of the dependencies in `docs/source/requirements.txt` to the following and still no success.

```
pyro-ppl==1.8.5
jupyter
mypy==1.5.0
sphinx==7.2.2
sphinxcontrib-bibtex==2.5.0
sphinx_rtd_theme==1.3.0
myst_parser==2.0.0
nbsphinx==0.9.2
```

eb8680:
By "no success," do you mean you're getting exactly the same error?

Issue #211, opened 2023-07-11
Update copyright to 2023


SamWitty:
https://basisresearch.github.io/chirho/getting_started.html

<img width="747" alt="Screenshot 2023-07-11 at 9 49 59 AM" src="https://github.com/BasisResearch/chirho/assets/13202375/3ae15469-6659-400e-a746-97478086e22d">

PR #255, opened 2023-09-07
Removed pandoc dependency from docs build CI

### LINKED ISSUE #210: Remove pandoc dependency in docs build
Building the documentation with `nbsphinx` currently requires `pandoc`, which I don't think can be installed through `pip`. I'm not sure exactly where it's being used, but we should look into it and remove the dependency if possible to further simplify installation.

SamWitty:
This tiny PR removes the unnecessary intallation of `pandoc` from the documentation CI build.

Closes #210

### DIFF
--- a/.github/workflows/publish_docs.yml
+++ b/.github/workflows/publish_docs.yml
@@ -15,10 +15,6 @@ jobs:
       pages: write
       id-token: write
     steps:
-    - id: pandoc-install
-      run: |
-        sudo apt install -y pandoc
-
     - id: deployment
       uses: sphinx-notes/pages@v3
       with:
--- a/.github/workflows/publish_docs.yml
+++ b/.github/workflows/publish_docs.yml
@@ -15,6 +15,9 @@ jobs:
       pages: write
       id-token: write
     steps:
+    - id: pandoc-install
+      run: |
+        sudo apt install -y pandoc
 
     - id: deployment
       uses: sphinx-notes/pages@v3
--- a/.github/workflows/publish_docs.yml
+++ b/.github/workflows/publish_docs.yml
@@ -15,9 +15,6 @@ jobs:
       pages: write
       id-token: write
     steps:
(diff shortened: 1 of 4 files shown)


SamWitty:
@eb8680 , this seems to be evidence that pandoc is necessary.

eb8680:
What I originally meant by #210 was to try to pinpoint where exactly `pandoc` was being used and see if those parts could be modified to remove the requirement, because I didn't think `pandoc` could be installed with `pip` alongside the other dependencies.

After some quick googling, it does look like there's a separate [PyPI `pypandoc` package `pypandoc_binary`](https://pypi.org/project/pypandoc/) that bundles a `pandoc` binary, though, so maybe all we need to do is add `pypandoc_binary` to the dependencies:
> Pypandoc uses pandoc, so it needs an available installation of pandoc. Pypandoc provides 2 packages, "pypandoc" and "pypandoc_binary", with the second one including pandoc out of the box. The 2 packages are identical, with the only difference being that one includes pandoc, while the other don't.

SamWitty:
Ok, I'll try that

eb8680:
Does using `pypandoc_binary` work for you locally? I guess it's OK to leave the manual installation step in the CI and docs deployment builds if `pypandoc_binary` works correctly in fresh local environments.

SamWitty:
Sort of. I'm not getting the `pandoc not installed` error that the CI builds in this PR are showing, but I am now getting the same doc building errors we're seeing in #243. What's perplexing to me is that the generic tests to non-master branches are successfully building the docs. E.g. See #211.

SamWitty:
I've now confirmed that when I `brew uninstall pandoc` and then install `pypandoc_binary` I get the same `pandoc missing` error we're seeing on these tests.

SamWitty:
Closing this WIP PR because the issue isn't as straightforward as I thought it would be. Will link to this for posterity in #210.

Issue #237, opened 2023-08-10
Add single-cell RNA seq example


agrawalraj:
We can look at this open-source project to easily load in single-cell RNA seq data in batches in pytorch:

https://chanzuckerberg.github.io/cellxgene-census/notebooks/experimental/pytorch.html

agrawalraj:
When I tried installing `cellxgene-census`, the `alias` dependency requires version 23.1.0 or above. I initially had an earlier version and got the error `TypeError: field() got an unexpected keyword argument 'alias'`. To use the PyTorch dataloader by importing `import cellxgene_census.experimental.ml as census_ml`, you need to install `torchdata`.

qinqian:
scvi has a more mature api for batch operation of single cell data with pytorch datasets. please see [here](https://github.com/scverse/scvi-tools/blob/1.0.3/scvi/data/_manager.py#L50-L533).

PR #265, opened 2023-09-14
Single-Cell RNA Seq Example

agrawalraj:
This example uses `ChiRho` to analyze a single-cell RNA seq experiment consisting of 188 drug perturbations on over 650k cells, and addresses #237. The raw data is downloaded from https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSM4150378 (the notebook automatically downloads and pre-processes the data by calling the script `SrivatsanTrapnell2020.py`).  

This notebook was done together with @qinqian and @eb8680!

**TODOs**:

- [x] Separate model and SVI code
- [x] Clean up guide specification with the blocking first line
- [x] Make Reporting regression coefficients section more approachable and more about fitting model. Might remove the boxplot of regression coefficients
- [x]  Write extended model for each query
- [x] Merge data cleaning script `SrivatsanTrapnell2020.py` into notebook
- [x] Add more description about the experiment and add Figure 1B and Figure 3A
- [x] Rewrite "Goal" section text and discussion of what we want to obtain from our analysis
- [x] Add more description about data prepping section and move after model rendering step
- [x] Rewrite text of new Fitting observed data section
- [x] For Computing causal queries section, want less causal inference terminology and write the actual estimand for Y_hat
- [x] Motivate why look at conditional average treatment effect than just average treatment effect
- [x] Update table of contents after section headings are finalized
- [x] Explicitly write causal estimands / make description flow more linear
- [x] Restrict to top 1000 variables genes + small set of target genes
- [x] Update figures to remove any mention of HDAC genes
- [x] Add figure 5 from paper 
- [x] For both ATE and CATE, redo summaries of scientific explanations. Have concluding paragraph of what we learned from the analysis
- [x] Rewrite last paragraph of Exploratory data analysis section


djinnome:
Note that HDAC inhibitors like Pracinostat affect the deacetylase activity of the HDAC protein.  What you are looking at is the gene expression of HDAC, which according to the central dogma would be upstream of the intervention.  

 If you want to estimate the causal effect of Pracinostat, look at transcription factors whose expression is [known to be suppressed by HDAC inhibition](https://en.wikipedia.org/wiki/Histone_deacetylase#HDAC_inhibitors), such as: [ACTR](https://en.wikipedia.org/wiki/Nuclear_receptor_coactivator_3), [cMyb](https://en.wikipedia.org/wiki/MYB_(gene)), E2F1, [EKLF](https://en.wikipedia.org/wiki/EKLF), [FEN 1](https://en.wikipedia.org/wiki/FEN-1), GATA, [HNF-4](https://en.wikipedia.org/wiki/Hnf4), HSP90, [Ku70](https://en.wikipedia.org/wiki/Ku70), NFκB, [PCNA](https://en.wikipedia.org/wiki/PCNA), p53, [RB](https://en.wikipedia.org/wiki/Retinoblastoma_protein), Runx, SF1 Sp3, STAT, [TFIIE](https://en.wikipedia.org/wiki/TFIIE), [TCF](https://en.wikipedia.org/wiki/TCF7L2), YY1.[[38]](https://en.wikipedia.org/wiki/Histone_deacetylase#cite_note-pmid15822187-38)[[39]](https://en.wikipedia.org/wiki/Histone_deacetylase#cite_note-pmid17694074-39), and the [downstream targets of those transcription factors](https://www.sciencedirect.com/science/article/abs/pii/S0300908421002534?via%3Dihub).

djinnome:
So the (simplified) causal diagram would be:

```
gene-> transcript -> protein -> HDAC activity -> transcription factor expression -> TF target expression
```

Where HDAC activity is the target of an HDAC inhibitor like Pracinostat, so the expression level of transcription factors would be expected to be suppressed by an HDAC inhibitor, but not HDAC gene expression.  Likewise, TF target expression could be downregulated or upregulated depending on whether the TF was an activator or inhibitor of the target.

eb8680:
@djinnome thanks, good points worth addressing in the example.

Keep in mind, though, that this is interventional data and that the scRNA-seq measurements are physically downstream of the interventions, since they are taken 24 hours after drug exposure. That means the causal effects in question are perfectly well-defined and identifiable from the data, which is the novelty and the promise of this kind of experiment.

The results in the notebook (and the [original paper](https://www.science.org/doi/10.1126/science.aax6234)) are consistent with your observations - HDAC enzyme activity is inhibited, so expression levels of some HDAC and SIRT genes are increased by cells in an attempt to offset the inhibition.

There is certainly confounding not taken into account here, especially in the form of selection bias (the experimental protocol and the preprocessing we do are both likely to filter out measurements of cells that died within 24 hours of drug treatment). A fuller version of this analysis would also attempt to jointly estimate treatment effects of all treatments on genome-wide gene expression, including all of the transcription factors you mentioned.

agrawalraj:
@djinnome thank you for the interesting points and explanations. In the notebook, we can add more description about 
the experimental setup (based on @eb8680 point) as to why the causal effects in question are well-defined.

qinqian:
Thanks @djinnome and @eb8680 for the in-depth discussion! We will definitely work on a fuller version of this example to address the prior knowledge of gene regulatory network and pathway information together.

eb8680 (approved):

PR #264, opened 2023-09-11
Adding undo_split and a test thereof (+small lint)

rfl-urbaniak:
`counterfactual.ops.py` now contains a definition of `undo_split`, which is tested at the bottom of `tests.indexed.test_internals.py`.

PR #268, opened 2023-09-19
Ru consequent differs

rfl-urbaniak:
Adding another helper function for AC. PR targets `ru-undo-split` and not `master`, so that explanation-related incremental steps lead to a single `explanation.py` and `test_handlers_explanation.py` in a linear manner, and can be conveniently put together before merging into `master`.


eb8680:
Blocked by #264, once that lands we can change the base branch to `master`.

eb8680 (approved):

Issue #269, opened 2023-09-19
Remove unused and empty `Operation` documentation header for `Counterfactual` 


SamWitty:
See https://basisresearch.github.io/chirho/counterfactual.html

<img width="981" alt="Screenshot 2023-09-19 at 12 33 32 PM" src="https://github.com/BasisResearch/chirho/assets/13202375/4667d504-ddee-47ca-ae0d-32f1b0600533">

PR #270, opened 2023-09-19
Removed unused headers for counterfactual operations

SamWitty:
addresses #269 

Building the docs locally with this change results in the following successfully built docs.
Screenshot 2023-09-19 at 12 46 14 PM


eb8680:
This seems like a bug, since `chirho.counterfactual.ops` is not empty.

SamWitty:
Yeah, good catch. I'll look into it.

PR #274, opened 2023-09-19
Make tutorial run end to end

### LINKED ISSUE #273: Tutorial notebook does not run end to end
I get the following error when running the notebook:

TypeError                                 Traceback (most recent call last)
Cell In[29], line 2
      1 from chirho.counterfactual.handlers import TwinWorldCounterfactual
----> 2 twin_model = TwinWorldCounterfactual(dim=-2)(bayesian_randomized_model)
      4 # Sample from both the factual and counterfactual worlds.
      5 twin_model(n_individuals)
TypeError: IndexPlatesMessenger.__init__() got an unexpected keyword argument 'dim'

@eb8680 fixed this error by (1) using `condition` from `chirho.observational.handlers` and fixing data type issues when using a `Bernoulli` distribution.

agrawalraj:
Resolves #273

- Uses `condition` from `chirho.observational.handlers`
- Removes `.long()` and instead calls `.int()` whenever we need to index into the conditional probability table from a random index sampled from a Bernoulli
- Makes everything a float (makes `smokes_assignment` into a float)


eb8680:
@SamWitty can you review this and double-check that the results are unchanged?

SamWitty (approved):
Looks good to me. Some plots changed in minor ways, but these changes are all within the realm of sampling differences. All qualitative conclusions hold exactly.

PR #278, opened 2023-09-20
Fix docstring propagation in counterfactual ops

### LINKED ISSUE #272: `@pyro.poutine.block` decorator causes documentation to fail to render
As briefly discussed (and misdiagnosed) in #269 and #270, the current docs build fails to render some operations. Some preliminary exploration points to this being a problem with the `@pyro.poutine.block` decorator, e.g. used in the `counterfactual.split` operations here https://github.com/BasisResearch/chirho/blob/master/chirho/counterfactual/ops.py#L13.

Currently, with the `@pyro.poutine.block` decorator the documentation renders as follows:
Screenshot 2023-09-19 at 2 26 23 PM

However, when I comment out only https://github.com/BasisResearch/chirho/blob/master/chirho/counterfactual/ops.py#L13 and run the docs build locally I get the following:
Screenshot 2023-09-19 at 2 27 08 PM

Note: getting the documentation to render does not require commenting out the `@pyro.poutine.runtime.effectful` decorator at https://github.com/BasisResearch/chirho/blob/master/chirho/counterfactual/ops.py#L12C1-L12C46. In addition, commenting out the `@pyro.poutine.runtime.effectful` decorator without commenting out the `@pyro.poutine.block` decorator does not fix the problem.

eb8680:
Resolves #272 

For whatever reason, calling `block` through `functools.partial` causes the docstrings of `split` and `preempt` to be passed through `block` correctly. Renders correctly on my machine.

### DIFF
--- a/chirho/counterfactual/ops.py
+++ b/chirho/counterfactual/ops.py
@@ -1,5 +1,6 @@
 from typing import Optional, Tuple, TypeVar
 
+import functools
 import pyro
 
 from chirho.indexed.ops import IndexSet, cond, scatter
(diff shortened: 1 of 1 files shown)


SamWitty (approved):
Confirmed that the docs build successfully on my machine.

Issue #288, opened 2023-09-22
Decouple `ODEDynamics` class from `TorchDiffEq`


SamWitty:
Currently, the definition of the `ODEDynamics` class assumes use of a `TorchDiffEq` backend by default, and is not modular 
with respect to other alternative backends. Instead, we should decouple the definition of an `ODEDynamics` model from the particular backend used to solve it.

PR #286, opened 2023-09-22
Refactor dynamical module file structure to be consistent with other modules

SamWitty:
Addresses #287 .

This refactoring PR moves dynamical system functionality in the `dynamical` module in to follow the following conventions.

`ops` -> User-facing classes and operations for representing dynamical systems, dynamical system state, and dynamical system trajectories.
`handlers` -> User-facing effect handlers for manipulating dynamical systems simulations.
`internals` -> Everything else.

A consequence of this refactoring is that any `ODE`-specific implementations have been moved into subdirectories of top-level `ops`, `handlers`, and `internals` directories.

PR #290, opened 2023-09-22
Decouple ODEDynamics class from TorchDiffEq

SamWitty:
Addresses #288 .

This refactoring PR introduces a dispatch mechanism for simulating a `ODEDynamics` model with a particular backend passed as a keyword argument to `simulate`, which is then passed to the newly introduced `ode_simulate` function. 

Other to-be-implemented backends now only need to provide implementations for `ode_simulate` and `ode_simulate_to_interruption`.

In addition, this PR introduces a convenience handler `BackendHandler` for inserting the backend argument into encapsulated `simulate` calls.


eb8680:
There are a bunch of merge conflicts to resolve following #286.

SamWitty:
Strange that there are all of these merge conflicts, as `sw-decouple-ODE` directly branched off of the `sw-dynamics-refactor-2` branch that was used in #286 . I suspect this is something to do with GitHub not recognizing squash-and-merge behavior? Maybe something else? Either way, I'll resolve by accepting all from `sw-decouple-ODE`.

chirho/dynamical/internals/ODE/backends/torchdiffeq.py
```diff
@@ -0,0 +1,288 @@
...
+        torch.transpose(
+            yt_[(..., None) + yt_.shape[len(yt_.shape) - event_dim :]],
+            -len(yt_.shape) - 1,
+            -1 - event_dim,
+        )[0]
+        for yt_ in yt_raw
+    )
+    return yt if event_fn is None else (event_t, yt)
+
+
```
agrawalraj, line 103: Code below is new; code above is old

chirho/dynamical/internals/ODE/backends/torchdiffeq.py
```diff
@@ -0,0 +1,288 @@
...
+@ode_simulate.register(TorchDiffEq)
```
agrawalraj, line 104: This is the key functionality / small indirection from before

chirho/dynamical/ops/ODE/backends/torchdiffeq.py
```diff
@@ -0,0 +1,15 @@
+from chirho.dynamical.ops.ODE import ODEBackend
+
+
+class TorchDiffEq(ODEBackend):
```
agrawalraj, line 4: This is a class so we can dispatch on it

agrawalraj (changes requested):
Could you also run the notebook to make sure the the inferences still make sense?

agrawalraj:
Otherwise, everything looks really great @SamWitty!

agrawalraj (approved):
Looks great!

PR #292, opened 2023-09-26
Consolidate `Backend` and `SolverHandler` into a single `Solver` effect handler / dispatch mechanism

SamWitty:
This refactoring PR primarily condenses what was previously two classes `Backend` and `SolverHandler` into a single `Solver` effect handler class. The key insight is that the `Solver` effect handler can both (i) be used to dispatch an implementation of `simulate` and (ii) be used as a context manager/effect handler to place itself as a `kwarg` in enclosing calls to `simulate`.

This also involves moving `Solver` classes that were previously in `ops` into `handlers`, resulting in a deceptively large number of file changes.

In addition, this PR renames `ODEBackend` to `ODESolver`, which was accidentally missed in #290 .


agrawalraj (approved):
Looks good to me!

agrawalraj (changes requested):
Looks good to me! Can you try linting again? The lint test is failing.

SamWitty:
> Looks good to me! Can you try linting again? The lint test is failing.

I can't tell what's happening here. The linter passes for me locally and the failed test is entirely uninformative. Can you reproduce locally?

agrawalraj:
> > Looks good to me! Can you try linting again? The lint test is failing.
> 
> I can't tell what's happening here. The linter passes for me locally (even with a fresh Python 3.8 environment) and the failed test is entirely uninformative. Can you reproduce locally?

I was able to reproduce error locally. I just committed the lint error. Hopefully that will make the tests pass here!

Issue #289, opened 2023-09-22
Decouple `SimulatorEventLoop` from concatenation of trajectories


SamWitty:
Currently, in the `staging-dynamics` branch the `SimulatorEventLoop` is responsible for tracking and executing interruptions, simulating the dynamical system between interruption events, and concatenating the results into a single trajectory. However, the concatenation of trajectories can/should be decoupled from managing interruptions.

To address this `simulate` should by default only return the final state of the trajectory, and thus the `SimulatorEventLoop` is by-default only responsible for producing the final state of the final simulation after all interruptions have been introduced.

We can/should then introduce a separate `Logging` effect handler for returning trajectories rather than the default behavior of returning the final state. `Logging` and `SimulatorEventLoop` should be modular and freely compose.

eb8680:
Dupe of #159?

SamWitty:
Yes, will close #159 in favor of this more descriptive issue.

SamWitty:
Blocked by #290

PR #293, opened 2023-09-27
Decouple `SimulatorEventLoop` from concatenation of trajectories

SamWitty:
Addresses #289 

Pair programmed with @agrawalraj .

This refactoring PR changes the default behavior of `simulate` to return a `State`, rather than a `Trajectory` as it was previously. Instead, we now introduce a `DynamicTrace` handler for extracting and storing whole trajectories from simulations. 

In order to implement `DynamicTrace`, a moderate refactoring of the `SimulatorEventLoop` was necessary. Specifically, `simulate_to_interruption` is now implemented in a way that is dynamical system agnostic, reducing to first determining when the interruption occurs and then recursively calling `simulate`. Each recursive `simulate` is then handled by various handlers.

`DynamicTrace` and `SimulatorEventLoop` commute.

Additional details:
- `NonInterruptionPointObservationArray` now subclasses `DynamicTrace`, substantially reducing LOC. Intuitively, this implementation simply generates a trajectory and then calls `dynamics.observation` on the simulated trajectory.
- Some classes and operations were minorly renamed to improve clarity. See Files Changed for details.


chirho/dynamical/internals/ODE/backends/torchdiffeq.py
```diff
@@ -106,75 +107,60 @@ def torchdiffeq_ode_simulate(
...
-
-    if nostat:
-        # This is required because torchdiffeq.odeint_event appears to just go on and on forever without a terminal
-        #  event.
-        raise ValueError(
-            "No static terminal interruption provided, but about to perform an event sim."
-        )
-    # for linter, because it's not deducing this from the if statement above.
-    assert next_static_interruption is not None
-
```
agrawalraj, line 155: All of this code not torchdiffeq specific moved into internals/dynamical/simulate_to_interruption.py

chirho/dynamical/internals/dynamical.py
```diff
@@ -15,28 +43,3 @@ def unsqueeze(x, axis: int):
 @unsqueeze.register
 def _unsqueeze_torch(x: torch.Tensor, axis: int) -> torch.Tensor:
     return torch.unsqueeze(x, axis)
-
-
-def _index_last_dim_with_mask(x: torch.Tensor, mask: torch.Tensor) -> torch.Tensor:
```
agrawalraj, line 20: This function got moved to ops

chirho/dynamical/internals/interruption.py
```diff
@@ -81,38 +122,3 @@ def apply_interruptions(
     """
     # Default is to do nothing.
     return dynamics, start_state
-
```
agrawalraj, line 84: We only append and do not concatenate anymore

agrawalraj (approved):
Looks fantastic! I really like how a lot of the logic got simplified during the refactor. In a later pass, I think we can refactor some of the tests and rename some of the handlers.

PR #294, opened 2023-10-03
Adding  helper function for generating stochastic interventions with approximately uniform distributions. 

### LINKED ISSUE #246: Add and test helper functions for generating uniform random interventions
The design in https://github.com/BasisResearch/chirho/pull/240 and previous iterations includes a number of new components on which the overall correctness of causation and explanation query implementations depends. To reduce the difficulty of reasoning about these implementations and make incremental progress toward landing causation and explanation queries in master, we should pull out these components and specify and test them in isolation.

One of these components is a pair of helper function for generating stochastic interventions whose values are drawn from (approximately) uniform distributions. These can be used with `chirho.counterfactual.ops.split` to cast the search for alternative antecedent values as an inference problem. In the draft design in #240 these functions are called `uniform_proposal` and `random_intervention`:
```python
def uniform_proposal(
    support: pyro.distributions.constraints.Constraint, **kwargs
) -> pyro.distributions.Distribution:
    ...

def random_intervention(
    support: Constraint,
    name: str,
    *,
    make_proposal: Callable[[Constraint], Distribution] = uniform_proposal,
) -> Callable[[torch.Tensor], torch.Tensor]:

    def _random_intervention(value: torch.Tensor) -> torch.Tensor:
        event_shape = value.shape[len(value.shape) - support.event_dim:]
        proposal_dist = make_proposal(support, event_shape=event_shape)
        return pyro.sample(name, proposal_dist)

    return _random_intervention
```
`uniform_proposal` should heuristically construct a high-entropy normalized proposal distribution whose `.support` attribute matches the given `pyro.distributions.constraints.Constraint` object, and should be extensible to new `Constraint` subtypes, perhaps via `functools.singledispatch`:
```python
@functools.singledispatch
def uniform_proposal(
    support: pyro.distributions.constraints.Constraint, **kwargs,
) -> pyro.distributions.Distribution:
    if support is pyro.distributions.constraints.real:
        return pyro.distributions.Normal(0., 1.).mask(False)
    elif support is pyro.distributions.constraints.boolean:
        return pyro.distributions.Bernoulli(logits=torch.zeros(()))
    else:
        tfm = pyro.distributions.transforms.biject_to(support)
        base = uniform_proposal(pyro.distributions.constraints.real, **kwargs)
        return pyro.distributions.TransformedDistribution(base, tfm)
```
This is similar in spirit to `pyro.distributions.ImproperUniform`, but `ImproperUniform` is not normalized and does not implement `sample()` and hence cannot be used as a proposal distribution.

rfl-urbaniak:
Resolves #246 .

A few points to note:

- In applications, I can imagine users willing to specify the params of the normal distribution or to specify that distribution with interval support should be uniform over that interval. I generalized the definitions accordingly, happy to discuss or revise.

-  Not sure if the shape or sample size parametrization would bring value to the tests of uniform_proposal, _uniform_proposal_indep, did not include at least for now.

-  Linting didn't recognize a registered function as used, manually added `noqa F401` to `test_handlers_explanation.py`.

- In the original definition for `_uniform_proposal_indep` this didn't work, as it asked for event shape from `kwargs` while it was already passed as a separate argument:

`return d.expand(kwargs["event_shape"]).to_event(support.reinterpreted_batch_ndims)`

Did I miss a subtlety here? Were two different event shapes meant to be passed here? For now replaced and tested with:

`return d.expand(event_shape).to_event(support.reinterpreted_batch_ndims)`

-  I'm unsure if calling the function `random_intervention` is the most transparent move, as the result is only a function performing `pyro.sample` over an appropriate uniform distribution. I would consider changing the name to `random_value` or `random_sample`.


eb8680:
@rfl-urbaniak it looks like you accidentally committed some log files, can you remove those?

rfl-urbaniak:
Removed the logs.

chirho/counterfactual/handlers/explanation.py
```diff
@@ -80,3 +82,152 @@ def _consequent_differs(consequent: T) -> torch.Tensor:
...
+    d = uniform_proposal(support.base_constraint, event_shape=event_shape, **kwargs)
+    return d.expand(event_shape).to_event(support.reinterpreted_batch_ndims)
+
+
+@uniform_proposal.register
+def _uniform_proposal_integer(
+    support: pyro.distributions.constraints.integer_interval,
+    **kwargs,
+) -> pyro.distributions.Distribution:
+    """
```
eb8680, line 168: You can remove the docstrings from registered implementations of `uniform_proposal` since they won't be rendered correctly by Sphinx and won't be visible through `uniform_proposal` in a REPL. All the documentation necessary for understanding `uniform_proposal` should go into the docstring of `uniform_proposal` itself.

rfl-urbaniak:
- Removed redundant parametrization from `uniform_proposal` and `random_intervention`
- Parametrized tests thereof, using `constraint.check`.
- Note that the positive constraint leads to a distribution that does not exclude zeros (so, strictly speaking,
fails the `constraint.check` test if those are not removed from the sample. I assume this is harmless.

do you want to keep the name random_intervention, as the function does not do any interventions?

eb8680 (approved):
Looks good to me, thanks!

PR #240, opened 2023-08-31
Attempt to modularize AC query code

eb8680:
Part of modularizing and simplifying #236 #164 
Targets #236, not `master`
Depends on #238 #239 

This PR is an attempt at simplifying the query code from @rfl-urbaniak's notebooks and turning it into library code. It attempts to be faithful to the notebooks, although after going through this exercise I have some new questions about the relationships between the textbook definitions and the latest query code in #236.

Only the changes in the new files `chirho/counterfactual/handlers/explanation.py` and `test/counterfactual/test_handlers_explanation.py` are new here. The others are merged from #238 and #239 and should land in `master` relatively soon.

PR #297, opened 2023-10-06
SearchForCause effect handler with tests thereof

### LINKED ISSUE #247: Add and test a handler for intervening on random subsets of variables
The design in https://github.com/BasisResearch/chirho/pull/240 and previous iterations includes a number of new components on which the overall correctness of causation and explanation query implementations depends. To reduce the difficulty of reasoning about these implementations and make incremental progress toward landing causation and explanation queries in `master`, we should pull out these components and specify and test them in isolation.

One of these components is an effect handler that intervenes on randomly selected subsets of a set of possible variables. In the draft design in #240 this handler is called `PartOfCause`:
```python
class PartOfCause(Protocol[P, S, T]):
    def __init__(self, actions: Mapping[str, Intervention[T]], *, bias: float, prefix: str): ...
    def __call__(self, fn: Callable[P, S]) -> Callable[P, S]: ...
    def __enter__(self): ...
    def __exit__(self, *args): ...
```
The implementation in #240 is quite compact thanks to #244's `undo_split` and the syntactic sugar of `contextlib.contextmanager`:
```python
@contextlib.contextmanager
def PartOfCause(
    actions: Mapping[str, Intervention[T]],
    *,
    bias: float = 0.0,
    prefix: str = "__cause_split_",
):
    preemptions = {
        antecedent: undo_split(antecedents=[antecedent])
        for antecedent in actions.keys()
    }

    with do(actions=actions):
        with BiasedPreemptions(actions=preemptions, bias=bias, prefix=prefix):
            with pyro.poutine.trace() as logging_tr:
                yield logging_tr.trace
```

rfl-urbaniak:
Resolves #247 

`PartOfCause` is now renamed as `SearchOfCause`. Reason: it is meant to be used to search for actual cause sets among the subsets of `actions`, not to search for nodes which are parts of such subsets. The latter is an additional task, and if we decide to add a handler achieving this, `PartOfCause` might be more suitable for its name. 

- The two layered test that succeeded with the tentative version in #240 only did so by a lucky accident. That version of `undo_split` contained a bug which accidentally yielded zeros, and the test expected a zero. The bug has since then been fixed, the test failure came back. 

- Another part of the problem was that with `undo_split` in its current shape, the antecedents used in the definition of witness preemptions should now be the ones used to introduce the split - that is, not the witness node itself (as its name is not even among its own indices), as in the old version of the test, but rather the antecedent nodes included in the actions. 

- It also turned out that it makes a difference where `pyro.poutine.trace` is collected. If it is collected at the time `SearchOfCause` is deployed, it simply does not record the changes resulting from the witness preemption. What needs to happen is that trace collection has to be next to the model itself to capture the effects of all the effect handlers. For this reason, `SearchOfCause` now yields nothing instead of a trace, as yielding a trace might encourage use that will lead to incorrect results when preemptions are stacked. 

Correct use of witness preemption construction and stacking of effect handlers is now illustrated in `test_SearchOfCause_two_layers` contained in `test_handlers_explanation.py`, which now succeeds.


rfl-urbaniak:
Screenshot from 2023-10-06 10-15-37
Make lint and make format pass locally, fail on GitHub, not sure what's that about.

SamWitty:
> Make lint and make format pass locally, fail on GitHub, not sure what's that about.

@rfl-urbaniak, check the version of `black` in your local environment. You can see in the Lint CI build `black` that it's using 23.9.1. I also encountered this problem, and updating the `black` version solved it.

eb8680 (dismissed):

eb8680 (approved):

PR #298, opened 2023-10-06
Remove `ODEDynamics` subclass and corresponding `simulate` indirection through `ode_simulate`

SamWitty:
Previously, the `dynamical` module include an explicit class hierarchy relating `Dynamics` and `ODEDynamics` that was used for type-directed dispatch of `simulate` operations. While this distinction may somewhat more fully communicate modeling assumptions to users, there was no explicit restriction placed on `ODEDynamics` that constrain any further than the base `Dynamics` class. As a result, the implementation included a substantial degree of indirection and cross-dependencies that was only justified by implicit and soft constraints on Class usage.

This PR removed the `ODEDynamics` subclass, and all corresponding redirection of the base `simulate` operation. 

This PR did involve a small amount of rearranging of files to address circular imports caused by the removal of the indirection through `ode_simulate`.

To address the concern that users will define `Dynamics` models that are out-of-scope of any particular solver we will solver-specific static or runtime checks as handlers as a separate contribution. See the subsumed and closed PR #295 for discussion.


eb8680 (approved):
I don't quite understand why the imports are so convoluted, but I'm just going to assume there's a reason and merge this.

PR #299, opened 2023-10-06
Migrate torchdiffeq dependency from "extras" to "install_requires"

SamWitty:
Currently, the `dynamics` module requires `torchdiffeq` as its only available backend. While it is unfortunate to add `torchdiffeq` as a core dependency for all of `chirho`, it seems necessary until #202 is addressed.

As the title states, this small PR migrates `torchdiffeq` dependency from "extras", which may or not be installed when a user installs `chirho`, to "install_requires", which will always be installed.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install .[test]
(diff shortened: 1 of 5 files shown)


SamWitty:
`extras` is a bit a blunt instrument currently. I'll add a "dynamics" option.

eb8680 (approved):

PR #301, opened 2023-10-06
Simplify State and Trajectory types

eb8680:
This refactoring PR simplifies the `State` and `Trajectory` types in `chiro.dynamical` and removes some dead code in those types that was never used.


SamWitty (approved):
Looks good to me. The local import trick is something I always forget about, and is good to have in the back pocket.

PR #302, opened 2023-10-06
Remove unused unsqueeze function

eb8680:
This refactoring PR removes a function `chirho.dynamical.internals.dynamical.unsqueeze` that was no longer used anywhere in the codebase.


SamWitty (approved):
Looks good. Always happy to delete unused code.

PR #303, opened 2023-10-06
Clean up Trajectory.append

eb8680:
This refactoring PR moves `Trajectory.append` into `chirho.dynamical.handlers.trace` and makes it slightly more generic.

### DIFF
--- a/chirho/dynamical/handlers/trace.py
+++ b/chirho/dynamical/handlers/trace.py
@@ -45,11 +45,14 @@ def append_tensor(prev_v: torch.Tensor, curr_v: torch.Tensor) -> torch.Tensor:
 
 
 class DynamicTrace(Generic[T], pyro.poutine.messenger.Messenger):
+    trace: Trajectory[T]
+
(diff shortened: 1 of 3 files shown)


chirho/dynamical/handlers/trace.py
```diff
@@ -9,7 +10,43 @@
 T = TypeVar("T")
 
 
+@functools.singledispatch
+def append(fst, rest: T) -> T:
```
SamWitty, line 14: Should this be `append(fst: T, rest: T) -> T:`?

SamWitty (changes requested):
One tiny comment. Otherwise looks great.

chirho/dynamical/handlers/trace.py
eb8680: No, that doesn't work with `singledispatch`, which expects registered types to be subtypes of the original type on the first argument.  It's an annoying aspect of the way `singledispatch` interacts with `mypy`.

SamWitty (approved):
lgtm

PR #304, opened 2023-10-06
Call DynamicTrace._reset on __enter__

eb8680:
This PR fixes what appears to be a bug in `DynamicTrace` - the trajectory is not reset on each new context entry.


SamWitty (approved):
lgtm

PR #307, opened 2023-10-10
Add default behavior for `X.t` with `torchdiffeq` solver.

SamWitty:
This small PR allows `Dynamics` `diff` methods to include arbitrary functions of time, denoted using `X.t` in the differential expressions.

The additional implementation in `solver/torchdiffeq.py` is copied exactly from @eb8680 's example integration with MIRA in  https://github.com/BasisResearch/chirho/blob/eb-dynamical-load-mira/chirho/dynamical/mira.py.

In addition to being nice default behavior, pushing handling of time into `ChiRho`'s `dynamical` module makes it easier to implement compiled symbolic rate laws in terms of time in the DARPA ASKEM program.

### DIFF
--- a/chirho/dynamical/internals/solver/torchdiffeq.py
+++ b/chirho/dynamical/internals/solver/torchdiffeq.py
@@ -30,6 +30,10 @@ def _deriv(
     env: State[torch.Tensor] = State()
     for var, value in zip(var_order, state):
         setattr(env, var, value)
+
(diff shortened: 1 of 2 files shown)


SamWitty:
Given that I removed `staging-dynamic` in the github CI for the PR to master in #241 , let's wait to resolve this until #241 lands.

eb8680 (approved):

Issue #160, opened 2023-06-21
Dynamical system runtime performance is slow! Quadratic in number of observations!


SamWitty:
Bonus: Try using JIT compilation. See (https://github.com/pyro-ppl/pyro/blob/727aff741e105715840bfdafee5bfeda7e8b65e8/pyro/infer/trace_elbo.py#L162)

https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html probably won't work with Pyro

SamWitty:
Addressed by `NonInterruptingPointObservationArray` introduced in #289 .

PR #309, opened 2023-10-11
Simplify SimulatorEventLoop logic

eb8680:
This refactoring PR simplifies the core `SimulatorEventLoop` and `Interruption` logic. It also fixes some type errors.

### DIFF
--- a/chirho/dynamical/handlers/dynamical.py
+++ b/chirho/dynamical/handlers/dynamical.py
@@ -15,7 +15,6 @@ T = TypeVar("T")
 
 
 class SimulatorEventLoop(Generic[T], pyro.poutine.messenger.Messenger):
-
     def _pyro_simulate(self, msg) -> None:
         dynamics, state, start_time, end_time = msg["args"]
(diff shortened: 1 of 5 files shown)


chirho/dynamical/handlers/dynamical.py
```diff
@@ -19,78 +15,36 @@
...
-
-            # Block dynamic interventions that have triggered and applied more than the specified number of times.
-            # This will prevent them from percolating up to the simulate_to_interruption execution.
-            with pyro.poutine.messenger.block_messengers(
-                lambda m: (
-                    isinstance(m, DynamicInterruption)
-                    and m.max_applications <= interruption_counts.get(m, 0)
-                )
-                or isinstance(m, SimulatorEventLoop)
+                lambda m: m is self or (isinstance(m, Interruption) and m.used)
```
eb8680, line 29: I plan to further simplify and modify this logic in a followup PR that should also address #160, but this is a good intermediate point.

SamWitty (approved):
Looks great!

PR #310, opened 2023-10-11
Rename Dynamics to InPlaceDynamics

eb8680:
As part of changing the interface to `simulate` to be functional, this refactoring PR renames the abstract interface type `Dynamics` to `InPlaceDynamics` and separates out the `observation` method into a subtype. It also removes a type variable in the `Dynamics` type which was not being used.

### DIFF
--- a/chirho/dynamical/ops/dynamical.py
+++ b/chirho/dynamical/ops/dynamical.py
@@ -2,7 +2,6 @@ import functools
 import numbers
 from typing import (
     TYPE_CHECKING,
-    Callable,
     FrozenSet,
     Generic,
     Optional,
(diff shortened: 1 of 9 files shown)


chirho/dynamical/ops/dynamical.py
```diff
@@ -107,18 +109,30 @@ def to_state(self) -> State[T]:
         return ret
 
 
+Dynamics = Callable[[State[S]], State[S]]
```
SamWitty, line 112: Is this still used anywhere?

SamWitty (changes requested):
Looks great. I like the distinction between `InPlaceDynamics` and `ObservableInPlaceDynamics` and the corresponding changes to the `StaticObservation` handler type.

Just one minor question about the use of the `Dynamics`.

chirho/dynamical/ops/dynamical.py
eb8680: I removed it

SamWitty (approved):
lgtm

PR #311, opened 2023-10-11
Rename NonInterruptingPointObservationArray to StaticBatchObservation

eb8680:
This refactoring PR renames the batched observation handler to something shorter and more consistent with the other interruption handlers.


SamWitty (approved):
lgtm

PR #312, opened 2023-10-11
Move dynamical backend interface into one file

eb8680:
This refactoring PR collects the existing backend interface components in `chirho.dynamical` and puts them in a single module. In followup PRs I will consolidate and simplify this interface.


SamWitty (approved):
Looks great. I particularly like the change to `get_solver`

PR #313, opened 2023-10-11
Remove unnecessary arguments from Interruptions

eb8680:
This refactoring PR simplifies the interfaces of the `Interruption` handlers, especially `DynamicInterruption`, by removing arguments that have fixed values or can be determined automatically.


tests/dynamical/test_dynamic_interventions.py
```diff
@@ -87,14 +87,10 @@ def test_nested_dynamic_intervention_causes_change(
...
                 event_f=get_state_reached_event_f(ts1),
                 intervention=is1,
-                var_order=init_state.var_order,
-                max_applications=1,
             ):
                 with DynamicIntervention(
                     event_f=get_state_reached_event_f(ts2),
                     intervention=is2,
-                    var_order=init_state.var_order,
-                    max_applications=1,
```
SamWitty, line 97: This is making me realize that #309 implicitly changed the semantics of `DynamicInterruption` by omitting the interruption the first time it's in the `terminal_interruptions`. In other words, we always assume `max_applications=1`.

This is fine by me, we'll just want to make sure we document it somewhere.

SamWitty (approved):
Approved. Left a comment about an implication of #309 that I missed, but I'm ok with the result.

PR #314, opened 2023-10-11
Remove obsolete script from dynamical test suite

eb8680:
This small PR removes a script in `tests/dynamical` that is never used in testing and is now out of date.


SamWitty (approved):
lgtm

PR #317, opened 2023-10-12
Make Trajectory methods use indexed ops

eb8680:
This refactoring PR removes most of the internals of `Trajectory` and reimplements `__len__` and `__getitem__` in terms of `chirho.indexed`.

I would like to remove `Trajectory` entirely in a followup refactoring PR, but this is a good intermediate stopping point for review.


chirho/dynamical/ops/dynamical.py
```diff
@@ -55,55 +56,41 @@ def __getattr__(self, __name: str) -> T:
...
+        assert len(key.shape) == 1 and key.shape[0] > 1  # DEBUG
+
+        if not key.any():  # DEBUG
+            return Trajectory()
+
+        name_to_dim = {k: f.dim - 1 for k, f in get_index_plates().items()}
+        name_to_dim["__time"] = -1
+        idx = IndexSet(__time={i for i in range(key.shape[0]) if key[i]})
+        # TODO support event_dim > 0
+        return gather(self, idx, event_dim=0, name_to_dim=name_to_dim)
```
SamWitty, line 88: I don't understand how `gather` works here when the dispatch for `Trajectory` is removed in `chirho/dynamical/internals/_utils.py`. The tests pass, so what am I missing?

chirho/dynamical/ops/dynamical.py
```diff
@@ -55,55 +56,41 @@ def __getattr__(self, __name: str) -> T:
...
+        if not self.keys:
+            return 0
+
+        name_to_dim = {k: f.dim - 1 for k, f in get_index_plates().items()}
+        name_to_dim["__time"] = -1
+        # TODO support event_dim > 0
+        return len(indices_of(self, event_dim=0, name_to_dim=name_to_dim)["__time"])
+
+    def __getitem__(self, key: torch.Tensor) -> "Trajectory[T]":
+        assert key.dtype == torch.bool
```
SamWitty, line 77: Is there a way to push this assertion into the type signature of `key`?

SamWitty (changes requested):
A few minor questions, mostly for my own understanding. Clarification would be helpful.

chirho/dynamical/ops/dynamical.py
eb8680: `Trajectory` is a subclass of `State`, so the implementations of `gather`, etc. for `State` are used for `Trajectory`.

chirho/dynamical/ops/dynamical.py
eb8680: No, but I am going to remove `Trajectory` (and hence this method) entirely, so it doesn't really matter.

chirho/dynamical/ops/dynamical.py
SamWitty: oh right. :)

SamWitty (approved):

PR #322, opened 2023-10-13
Remove var_order attribute from State interface

eb8680:
This small refactoring PR removes the `State.var_order` attribute (which is an implementation detail of `torchdiffeq`) from the user-facing `State` type and moves it to `chirho.dynamical.internals._utils`.


SamWitty (approved):
Looks great. The functools memoization is a good tool in the toolbox.

Issue #321, opened 2023-10-12
Dynamical system Trajectory type is redundant


eb8680:
The `Trajectory` type in `chirho.dynamical` #241 is redundant with respect to the `State` type and `chirho.indexed` and should be removed before #241 lands on `master`.

eb8680:
Resolved by #327

PR #323, opened 2023-10-13
Remove usage of Trajectory.__len__ 

eb8680:
As part of addressing #321, this refactoring PR removes the `Trajectory.__len__` method. This is only used in one place, so I moved the existing logic there, and even that should not be necessary once the interaction of `LogTrajectory` and `InterruptionEventLoop` is cleaned up later.

### DIFF
--- a/chirho/dynamical/handlers/interruption.py
+++ b/chirho/dynamical/handlers/interruption.py
@@ -7,6 +7,7 @@ import torch
 
 from chirho.dynamical.handlers.trajectory import LogTrajectory
 from chirho.dynamical.ops import ObservableInPlaceDynamics, State
+from chirho.indexed.ops import get_index_plates, indices_of
 from chirho.interventional.ops import Intervention, intervene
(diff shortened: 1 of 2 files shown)


chirho/dynamical/handlers/interruption.py
```diff
@@ -170,22 +172,23 @@ def __init__(
         super().__init__(times)
 
     def _pyro_post_simulate(self, msg) -> None:
-        dynamics: ObservableInPlaceDynamics[torch.Tensor] = msg["args"][0]
-
-        if "in_SEL" not in msg.keys():
-            msg["in_SEL"] = False
+        super()._pyro_post_simulate(msg)
```
eb8680, line 175: I believe the previous location of this line was a silent bug.

eb8680: Fix included in #324

SamWitty (approved):
lgtm

PR #324, opened 2023-10-13
Fix generic types and arguments of LogTrajectory and StaticBatchObservation

eb8680:
This refactoring PR includes a few minor fixes for the interfaces of `LogTrajectory` and `StaticBatchObservation` to make them consistent with each other and the rest of the module.


SamWitty (approved):
lgtm

PR #325, opened 2023-10-13
Remove Trajectory.__getitem__ method

eb8680:
~~Blocked by #323~~

As part of addressing #321, this refactoring PR removes the `__getitem__` method of the `Trajectory` type, replacing it with a call to `chirho.indexed.ops.gather` as in the current implementation of `__getitem__`. This method was only used in two places in the code and a couple of unit tests, so the change isn't actually too big.

### DIFF
--- a/chirho/dynamical/handlers/interruption.py
+++ b/chirho/dynamical/handlers/interruption.py
@@ -7,6 +7,7 @@ import torch
 
 from chirho.dynamical.handlers.trajectory import LogTrajectory
 from chirho.dynamical.ops import ObservableInPlaceDynamics, State
+from chirho.indexed.ops import get_index_plates, indices_of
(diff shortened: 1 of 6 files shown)


eb8680:
Base branch should be changed back to `staging-dynamic` after #323 lands

SamWitty:
@eb8680 , could you please resolve merge conflicts here. Normally I would do it, but the conflicts in `trajectory.py` aren't entirely trivial and I don't want to accidentally revert changes in this PR.

SamWitty (approved):
Looks good to me. It might be nice to write a helper function to reduce duplicated code between `trajectory.py` and `torchdiffeq.py`, but that can be done in a subsequent PR. I understand that this is in progress towards pulling all methods out of the `Trajectory` class, which has some minor proliferation of duplicate code.

PR #329, opened 2023-10-13
Automated doubly robust inference

agrawalraj:
**TODOs**:
- [ ] Write simple design document and then make staging branch. After this step, we can make the TODOs below into issues. We should be able to easily write many other papers that use influence functions (e.g., black box influence function paper) in our framework
- [ ] Separate out efficient influence function computation from `one_step_correction`. `one_step_correction` should just be a simple call to the efficient influence function. This will allow us to easily define other procedures such as TMLE
- [x] Vectorize efficient influence function computation over multiple test points
- [ ] General refactoring of interfaces (might have some tensor indexing issues to fix here). For example, remove out monte_carlo plate from the model)
- [ ] Specify key assumptions on functionals (e.g., DICE assumes its a Pyro model or if we make linearity assumptions, the variance of `plug_in_grads` will reduce)
- [ ] Implement the Bayesian double robust version using efficient influence function function (maybe if you're doing variational inference, you work in terms of variational params instead of samples). Or might do some sort of adaptive importance sampling
- [ ] Extension to functionals that map to space of functions instead of finite-dimensional vectors (e.g., conditional average treatment effect)
- [ ] Adding general cross-fitting approach as in the double ML paper 
- [ ] Write many more unit tests (use Kennedy's influence function calculus to make powerful tests)
- [ ] Move cut posterior modules into "robust" once we finish the above TODOs and merge into master
- [ ] Set pre-deadline deadline for a potential paper submission (discuss if involves external collaborators)
- have collaborator (have really good example where it works like dr_learner notebook)
- by ourself 
- things aren't far enough long so push back deadline (behind schedule on wrapping up code or other pressing demands)
- Don't write paper (probably unlikely)


SamWitty:
@agrawalraj , I would recommend setting up a staging branch similar to `staging-dynamic`, as this will involve a fair amount of development work that we don't necessarily want on `main` until it's ready for release.

eb8680:
Closing in favor of #384

PR #330, opened 2023-10-16
Separate Observable from InPlaceDynamics interface

eb8680:
This small refactoring PR makes the `.observation` and `Dynamics` interfaces of `ObservableInPlaceDynamics` orthogonal, as a preliminary to possibly making the `Dynamics` interface functional by default.

### DIFF
--- a/chirho/dynamical/handlers/interruption.py
+++ b/chirho/dynamical/handlers/interruption.py
@@ -6,7 +6,7 @@ import pyro
 import torch
 
 from chirho.dynamical.handlers.trajectory import LogTrajectory
-from chirho.dynamical.ops import ObservableInPlaceDynamics, State
+from chirho.dynamical.ops import Observable, State
 from chirho.indexed.ops import get_index_plates, indices_of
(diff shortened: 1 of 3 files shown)


SamWitty (changes requested):
This looks good to me. I believe this PR should also change the `SIR` model in the `dynamic_fixtures` for dynamical tests to be a mixin of `InPlaceDynamics` and `Observable`. This likely slipped through the cracks of _this_ PR becaue the `SIR` model is just an instance of `InPlaceDynamics`, even though it has an `observation` method and should have been an instance of `ObservableInPlaceDynamics` in a previous PR.

eb8680:
> I believe this PR should also change the SIR model in the dynamic_fixtures for dynamical tests to be a mixin of `InPlaceDynamics` and `Observable`

`InPlaceDynamics` and `Observable` are `Protocol`s, so they don't need to be inherited from at all. I've removed `InPlaceDynamics` from the definition of `SIR` to illustrate.

SamWitty (approved):
lgtm

PR #331, opened 2023-10-16
Move keys property of State into a helper function get_keys

eb8680:
This refactoring PR makes the interface of `State` smaller and easier to change later by moving `.keys` from a `property` method to a separate function `get_keys`.


SamWitty (approved):
lgtm

SamWitty:
@eb8680 . Could you please resolve these merge conflicts. I think I could do it, but these are involved enough that I'd rather not make a mistake and have it slip through our review process.

SamWitty (approved):

eb8680:
Sure, done

PR #332, opened 2023-10-16
One-sided runtime check of whether a `dynamics` instance satisfies assumptions of a `solver`

SamWitty:
Dynamical systems solvers often make assumptions about the form the dynamical system. For example, ODE solvers like torchdiffeq assume that calls to `dynamics.diff` are fully determined by `state`. 

This small PR provides the scaffolding for implementing runtime checks on dynamical systems, and simple implementation of such a check for the `torchdiffeq` backend.


SamWitty:
Note: Discussed w/ @eb8680 and we decided that this torchdiffeq test should just check whether any enclosing `simulate` calls contain `sample` calls, and fail if so. This will involve having the `RuntimeCheckDynamics` handler insert another handler (specific to the backend) that checks the contents of other operations in the `simulate` call. For example, `torchdiffeq` should error when any `pyro.sample` statements are made within the `simulate` call.

eb8680:
Let's get #361 merged first, since it changes some of the interfaces used here

SamWitty:
@eb8680 , this is now much simpler and less invasive. Take a quick look again?

eb8680 (approved):
Looks good with latest simplifications

chirho/dynamical/handlers/event_loop.py
```diff
@@ -28,6 +29,9 @@ class InterruptionEventLoop(Generic[T], pyro.poutine.messenger.Messenger):
     def _pyro_simulate(msg) -> None:
         dynamics, state, start_time, end_time = msg["args"]
 
+        if pyro.settings.get("validate_dynamics"):
+            check_dynamics(dynamics, state, start_time, end_time, **msg["kwargs"])
```
eb8680, line 33: @SamWitty I went ahead and added the validation flag to this PR instead of making a new one.

SamWitty: @eb8680, while we're waiting for #380 to land, could you add this `check_dynamics` call to `simulate` as well?

PR #340, opened 2023-10-18
Use observe operation in dynamical system observation handlers

### LINKED ISSUE #334: Dynamical systems observation interface should use observe
As of #330, the interface for observations in `chirho.dynamical` is a class with an `observation` method corresponding to a generative model of observations:
```py
class Observable(Protocol[T]):
    def observation(self, state: State[T]) -> None: ...
```
The `chirho.dynamical` observation handlers `StaticPointObservation` and `StaticBatchObservation` both use this interface together with a `chirho.observational.condition` handler to constrain the `sample` statements inside `observation`:
```py
class StaticPointObservation(...):
    data: Dict[str, Observation[T]]

    def _pyro_apply_interruptions(self, msg):
        ...
        dynamics, state = msg["args"][0], msg["args"][1]
        with chirho.observational.handlers.condition(data=self.data):
            dynamics.observation(state)
```
Instead of this custom interface, we should have the `chirho.dynamical` observation handlers invoke `chirho.observational.ops.observe` directly on a dependent observation function corresponding to the current `observation` method plus the `condition` handler:
```py
class StaticPointObservation(...):
    observation: Callable[[State[T]], State[T]]

    def _pyro_apply_interruptions(self, msg):
        ...
        dynamics, state = msg["args"][0], msg["args"][1]
        obs_state = chirho.observational.ops.observe(state, self.observation)
        msg["value"] = (dynamics, obs_state)

...

def my_obs(state):  # wrapping the old interface to illustrate
    my_dynamics.observation(state)
    return state


h = StaticPointObservation(time=t, observation=condition(data=my_data)(my_obs))
```
This change would further simplify the interface of `chirho.dynamical` and make the observation handlers more consistent with the intervention handlers, which directly invoke `chirho.interventional.ops.intervene` to handle `apply_interruptions`.

eb8680:
Resolves #334 

This refactoring PR makes the handlers `chirho.dynamical.handlers.interruption.StaticObservation` and `chirho.dynamical.handlers.interruption.StaticBatchObservation` use the `chirho.observational.ops.observe` primitive directly, rather than assuming a model comes with a special `.observation` method.

### DIFF
--- a/chirho/dynamical/handlers/interruption.py
+++ b/chirho/dynamical/handlers/interruption.py
@@ -1,16 +1,15 @@
 import numbers
 import warnings
-from typing import Callable, Dict, Generic, Optional, TypeVar, Union
+from typing import Callable, Generic, Optional, TypeVar, Union
 
 import pyro
 import torch
 
 from chirho.dynamical.handlers.trajectory import LogTrajectory
-from chirho.dynamical.ops import Observable, State, get_keys
+from chirho.dynamical.ops import State, get_keys
 from chirho.indexed.ops import get_index_plates, indices_of
 from chirho.interventional.ops import Intervention, intervene
-from chirho.observational.handlers import condition
-from chirho.observational.ops import Observation
+from chirho.observational.ops import Observation, observe
 
 R = Union[numbers.Real, torch.Tensor]
 S = TypeVar("S")
@@ -86,15 +85,13 @@ class _InterventionMixin(Generic[T]):
 
 
 class _PointObservationMixin(Generic[T]):
-    data: Dict[str, Observation[T]]
+    observation: Observation[State[T]]
     time: R
 
     def _pyro_apply_interruptions(self, msg) -> None:
-        dynamics: Observable[T] = msg["args"][0]
-        current_state: State[T] = msg["args"][1]
-
-        with condition(data=self.data):
-            dynamics.observation(current_state)
+        dynamics = msg["args"][0]
+        state: State[T] = msg["args"][1]
+        msg["value"] = (dynamics, observe(state, self.observation))
 
     def _pyro_sample(self, msg):
         # modify observed site names to handle multiple time points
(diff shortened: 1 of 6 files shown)


chirho/dynamical/handlers/interruption.py
```diff
@@ -86,15 +85,13 @@ def _pyro_apply_interruptions(self, msg) -> None:
...
 
     def _pyro_apply_interruptions(self, msg) -> None:
-        dynamics: Observable[T] = msg["args"][0]
-        current_state: State[T] = msg["args"][1]
-
-        with condition(data=self.data):
-            dynamics.observation(current_state)
+        dynamics = msg["args"][0]
+        state: State[T] = msg["args"][1]
+        msg["value"] = (dynamics, observe(state, self.observation))
```
SamWitty, line 94: Would we ever expect `observe` to change its output to anything other than `state`? If it's just a side effect that changes the likelihood then maybe the following would be better?

```
state: State[T] = msg["args"][1]
observe(state, self.observation)
```

chirho/dynamical/handlers/interruption.py
```diff
@@ -177,6 +174,4 @@ def _pyro_post_simulate(self, msg) -> None:
         )
 
         if len_traj == len(self.times):
-            dynamics: Observable[T] = msg["args"][0]
-            with condition(data=self.data):
-                dynamics.observation(self.trajectory)
+            msg["value"] = observe(self.trajectory, self.observation)
```
SamWitty, line 177: See comment in `_PointObservationMixin`

SamWitty (changes requested):
Overall looks good. I have a few minor questions that should be easy to address.

chirho/dynamical/handlers/interruption.py
eb8680: Possibly, yes - consider a particle filter that updates particles at each observe() call

SamWitty:
@eb8680 , looks like you've addressed all my comments. Is this ready for review again?

SamWitty (approved):
lgtm

PR #341, opened 2023-10-18
Use functional interface for dynamical systems

### LINKED ISSUE #335: Primary dynamical systems interface should be functional rather than in-place
As of #330, the main `InPlaceDynamics` interface used throughout `chirho.dynamical` uses mutable state as a destination and expects a specially named `diff` method:
```py
class InPlaceDynamics(Protocol[T]):
    def diff(self, state: State[T], dstate: State[T]) -> None: ...

def simulate(dynamics: InPlaceDynamics[T], ...): ...
```
Once #334 is addressed, this could be simplified to a functional interface with no special method name required:
```py
Dynamics = Callable[[State[T]], State[T]]

def simulate(dynamics: Dynamics[T], ...): ...
```
We could then define syntactic sugar like `InPlaceDynamics` on top of this minimal core interface.

eb8680:
Blocked by #340 

Resolves #335

This refactoring PR removes the `InPlaceDynamics` interface from #330 with its special `.diff` method and replaces it with a simpler generic functional interface.

After this change, `chirho.dynamical.ops.simulate` just takes a generic function `dynamics: Callable[[State[T]], State[T]]` that takes a state vector and returns a difference or derivative vector as its model argument.

### DIFF
--- a/chirho/dynamical/handlers/interruption.py
+++ b/chirho/dynamical/handlers/interruption.py
@@ -1,16 +1,15 @@
 import numbers
 import warnings
-from typing import Callable, Dict, Generic, Optional, TypeVar, Union
+from typing import Callable, Generic, Optional, TypeVar, Union
 
 import pyro
 import torch
 
 from chirho.dynamical.handlers.trajectory import LogTrajectory
-from chirho.dynamical.ops import Observable, State, get_keys
+from chirho.dynamical.ops import State, get_keys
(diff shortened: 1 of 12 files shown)


eb8680:
Base branch should be changed back to `staging-dynamic` after #340 is merged.

SamWitty (approved):
Looks good to me.

PR #349, opened 2023-10-18
Remove get_keys operation

eb8680:
After #346 changed `State` to a vanilla `dict`, the `get_keys` operation added in #331 is no longer necessary. This refactoring PR removes `get_keys` and replaces it with `dict.keys`.


SamWitty (approved):
lgtm

PR #354, opened 2023-10-19
Remove InterruptionEventLoop and other refactorings...

SamWitty:
This larger draft PR is intended to facilitate discussion about the ongoing design efforts for the `dynamical` module. Once we reach a consensus around this next phase of design, this will be broken up into substantially smaller PRs.

PR #360, opened 2023-10-25
Add shallow effect handler base class

eb8680:
Part of addressing performance and design issues related to #354 #160 

This PR adds an abstract "shallow" effect handler class `ShallowMessenger` to `chirho.dynamical.internals._utils` that exits after handling a single operation call. In a followup PR, I will use `ShallowMessenger` as a base class for `chirho.dynamical.handlers.interruption.Interruption`.


SamWitty (approved):
lgtm

PR #362, opened 2023-10-27
adding ExplainCauses

rfl-urbaniak:
This adds the `ExplainCauses` handler used for causal explanation search and a test thereof.


eb8680 (approved):
Looks great! I think the docs could use some work, but let's do that separately.

PR #364, opened 2023-10-31
Fixed SDID bug 

### LINKED ISSUE #353: Issue in SDID notebook
@rfl-urbaniak tried to use apply this notebook for cities project, but ran into tensor shape issues. I think the issue is that this line is incorrect:

```
with pyro.plate("time_weights", self.N):
    pyro.sample("avg_y_post_treat", dist.Normal(y_time, 1.0), obs=self.avg_y_post_treat)
```

Based on formula 2.3 in the synthetic DID paper, this plate should only be over the number of control units and the shapes of `self.avg_y_post_treat` and `self.avg_y_post_treat` are wrong. The notebook only ran because `self.avg_y_post_treat` was a vector of length 1 (since there's only a single treated unit) which allowed for broadcasting. The notebook also makes implicit assumptions about the way the data is formatted (e.g., that all the control units have a `unit_index` that those of treated units). We should make the code more generalizable to other data formats.

agrawalraj:
Resolves #353.

I also added an assert statement in `BayesSDID` that checks if the input dataframe is formatted correctly.


rfl-urbaniak (approved):
Looks good, thanks!

Issue #350, opened 2023-10-18
`chirho.dynamical.ops.State` should just be a type alias


eb8680:
After #346, `State` is a generic subclass of `dict`:
```py
class State(Generic[T], dict[str, T]):
  pass
```
The final step in removing the unnecessary custom type `State` would be to make `State` a type alias:
```py
State = dict[str, T]
```
This will require moving `State`-specific patterns for `chirho.indexed` ops registered in `chirho.dynamical._utils` to generic `dict` patterns registered in `chirho.indexed.internals`.

Tasks:
- [x] #374 
- [x] #375 
- [ ] Make `State` a `Mapping` alias and remove `State()` constructor calls #433

PR #374, opened 2023-11-13
Separate cond_n and scatter_n from cond and scatter

eb8680:
Addresses #350 

This refactoring PR promotes `cond_n` and `scatter_n` to separate operations in `chirho.ops.indexed`, rather than patterns of `cond` and `scatter`.

This change is necessary for #350 because reducing `dynamics.ops.State` to a type alias of `dict` would require new patterns for `dict` that would collide with the patterns corresponding to `cond_n` and `scatter_n`, which don't really make sense as specializations of the binary operations `cond` and `scatter` anyway.


SamWitty (approved):
This looks reasonable to me.

PR #375, opened 2023-11-13
Make State a type alias rather than a subtype

eb8680:
Addresses #350. Blocked by #374.

This refactoring PR makes `chirho.dynamical.ops.State` into a type alias of `dict[str, T]`.

### DIFF
--- a/chirho/dynamical/internals/_utils.py
+++ b/chirho/dynamical/internals/_utils.py
@@ -6,39 +6,9 @@ import pyro
 import torch
 
 from chirho.dynamical.ops import State
-from chirho.indexed.ops import IndexSet, gather, indices_of, union
-from chirho.interventional.handlers import intervene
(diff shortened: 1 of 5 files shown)


SamWitty (changes requested):
This all looks reasonable. That being said, the tests currently fail because the type-alias `State` is being used to instantiate throughout the codebase and in the tests.

I quickly tried to just fix this myself by find-replacing all instances of `State(` with `dict(`, but I then get the following error:

`TypeError: Subscripted generics cannot be used with class and instance checks`

See https://github.com/BasisResearch/chirho/compare/master...sw-revise-tests-state-alias for the changes I attempted.

eb8680:
> I quickly tried to just fix this myself by find-replacing all instances of State( with dict(

I fixed this error with a less invasive but more hacky solution, but this is probably still worth doing in a follow-up PR.

Issue #344, opened 2023-10-18
Consider splitting up dynamical system tutorial


eb8680:
The current dynamical systems tutorial notebook in #241 is quite long. Maybe it would be easier to digest if it were broken up into 2-3 smaller tutorials covering different segments of the existing content?

SamWitty:
I think this is a good idea. @agrawalraj and I were also talking about adding two additional example notebooks that demonstrate alternative usage of the dynamical systems functionality. E.g. PINNs and NeuralODEs. I'll make separate issues for those. The key difference would be the usage of neural networks for representing the solution (as I believe in PINNs) or the dynamical system itself (as I believe as in NeuralODEs).

My suggestion for this notebook, which I'd describe as "causal and statistical reasoning with mechanistic dynamical systems", would be to separate it out into 3 notebooks.

1. Interventional reasoning with mechanistic dynamical systems
2. Counterfactuals and actual causality with mechanistic dynamical systems
3. Bayesian hierarchical multilevel models with mechanistic dynamical systems

Issue #343, opened 2023-10-18
Dynamical systems tutorial is missing textual content


eb8680:
The introductory notebook `dynamical_intro.ipynb` in #241 needs explanatory text to be written for most of the code and results.

eb8680:
Resolved by #377

PR #377, opened 2023-11-13
Revise basic SIR tutorial

SamWitty:
This PR revises the basic example as follows:

1. Adds textual content, addressing #343.
2. Changes the plots to be more short-term focussed, satisfying the request of DARPA ASKEM program manager.
3. Removes the counterfactual reasoning with dynamical systems, as discussed in #344 and subsumed by #376.
4. Suppresses deprecated usage warnings from `seaborn` 
5. Adds the notebook (with smoke test parameters) to the CI partially addressing #410


SamWitty:
> Looks much better! Two comments:
> 
> * Can you remove the `.diff` method indirection in the models following [Use functional interface for dynamical systems #341](https://github.com/BasisResearch/chirho/pull/341)?
> * Should we consider moving the `pyro.deterministic` calls in helper functions like `simulated_bayesian_sir` that expose the trajectory to `Predictive` into our `LogTrajectory` handler?

@eb8680 , I believe these are now addressed.

PR #382, opened 2023-11-16
Slime mold dynamical system model

emackev:
A demo notebook to:
- Write a (simple 1D) dynamical systems model of slime mold
- Simulate the slime mold model
    - from known parameters
    - from a Bayesian model with unknown parameters and measurement noise
- Use ChiRho to infer the model parameters from data
- Use ChiRho to infer what would happen if we intervene


SamWitty:
@emackev , are you still interested in turning this notebook into a full-blown example like https://basisresearch.github.io/chirho/backdoor.html or https://basisresearch.github.io/chirho/mediation.html?

emackev:
> @emackev , are you still interested in turning this notebook into a full-blown example like https://basisresearch.github.io/chirho/backdoor.html or https://basisresearch.github.io/chirho/mediation.html?

Yes! What's the best way to go about this? Should it be a stand-along thing, or combined with another demo?

SamWitty:
I think stand-alone thing is best, to keep the size of each tutorial relatively small. I'll rearrange the dynamical systems tutorials into a coherent narrative together after we have them all in place.

It seems like the main thing that's missing is: (1) explanatory text, (2) a "success" story (which this might be already due to non-identification, we'd just need to explore the posterior predictive distributions), and (3) ideally real or real-ish data.

PR #384, opened 2023-11-22
Staging branch for `chirho.robust` development

eb8680:
Closed in favor of #398


chirho/robust/ops.py
```diff
@@ -0,0 +1,48 @@
...
+    # TODO check that target_params == model_params | guide_params
+    assert isinstance(target, torch.nn.Module)
+    target_params, func_target = make_functional_call(target)
+
+    @functools.wraps(target)
+    def _fn(point: Point[T], *args: P.args, **kwargs: P.kwargs) -> S:
+        param_eif = linearized(point, *args, **kwargs)
+        return torch.func.jvp(
+            lambda p: func_target(p, *args, **kwargs), (target_params,), (param_eif,)
+        )[1]
```
agrawalraj, line 46: If `param_eif` is batched over multiple points, then we should replace this line with something like this:

```
return torch.vmap(lambda d: torch.func.jvp(lambda p: func_target(p, *args, **kwargs), (target_params,), (d,))[1], in_dims=0)(param_eif)
```

agrawalraj (changes requested):
It's really coming together! Thanks for organizing everything so nicely :)

PR #395, opened 2023-11-28
Merge InterruptionEventLoop into Solver

### LINKED ISSUE #380: InterruptionEventLoop should be absorbed into Solver
Currently, evaluating `chirho.dynamical.ops.simulate` requires explicitly creating and applying two separate effect handlers, a `Solver` and an `InterruptionEventLoop`. `simulate` also accepts a `Solver` handler as a keyword argument, an unusual and non-intuitive syntactic pattern for an effectful operation.

#361 cleared the remaining obstacles to eliminating this verbose and non-standard syntax through combining these two handlers into one base `Solver` handler that can be shared across backends.

Tasks:
- [x] #361
- [ ] Merge `InterruptionEventLoop` and `Solver` #395 
- [ ] Remove the `solver` keyword argument from `simulate` #395

eb8680:
Resolves #380 

This refactoring PR removes `InterruptionEventLoop` and moves its functionality into the base `Solver` handler in `chirho.dynamical.internals.solver`. It also removes the `solver` keyword argument from `simulate`.


chirho/dynamical/handlers/solver.py
```diff
@@ -1,7 +1,16 @@
+from __future__ import annotations
+
+from typing import TypeVar
+
+import torch
+
 from chirho.dynamical.internals.solver import Solver
 
+S = TypeVar("S")
```
SamWitty, line 9: Looks like these TypeVars aren't used anymore.

SamWitty (changes requested):
Looks good. One minor nit about TypeVars.

chirho/dynamical/handlers/solver.py
eb8680: There are lots of those in the codebase, but neither `mypy` nor `black`/`flake8` seem to mind so I just let them be.

SamWitty (approved):

Issue #396, opened 2023-11-28
LogTrajectory does not handle edge cases correctly


eb8680:
The logging handler `chirho.dynamical.handlers.LogTrajectory` does not correctly handle edge cases where the `start_time` of a simulation interval (including an interruption time) is one of the logging times.

Minimal failing example:
```python
def test_mwe_fail():
    from chirho.dynamical.ops import State, simulate
    from chirho.dynamical.handlers.solver import TorchDiffEq
    from chirho.dynamical.handlers.trajectory import LogTrajectory

    dynamics = lambda s: State(X=s["X"] * (1 - s["X"]))
    init_state = State(X=torch.tensor(0.5))
    start_time, end_time = torch.tensor(0.), torch.tensor(3.)

    with TorchDiffEq(), LogTrajectory(times=torch.tensor([0., 1., 2.])) as log:
        final_state = simulate(dynamics, init_state, start_time, end_time)

    assert len(log.trajectory["X"]) == len(log.times) == 3  # fails bc len(X) == 2
```

SamWitty:
I'll take this on, as my previous PR (#363) was supposed to fix this.

PR #397, opened 2023-11-28
Fix time collision bug in `LogTrajectory`

SamWitty:
This PR addresses #396 by adding the initial state to the trajectory if the `start_time` of `simulate` is equal to the first element of the (sorted) `logging_times` argument of the `LogTrajectory` handler. This works because the `LogTrajectory` handler excludes time collisions on the `start_time` and includes time collisions on the `end_time` of each `simulate_point` call, which was necessary to not "double include" interruption times that collided with an element in the `logging_times` and thus show up as the `start_time` and `end_time` of two `simulate_point` calls.

In addition, this PR adds a slight modification of the test described in #396 to the test suite.


chirho/dynamical/handlers/trajectory.py
```diff
@@ -27,6 +31,19 @@ def __enter__(self) -> "LogTrajectory[T]":
         self.trajectory: State[T] = State()
```
SamWitty, line 31: @eb8680 , do you think it makes sense to move this line into `_pyro_simulate` instead of `__enter__`. This would change the behavior from "concatenate multiple `simulate` calls trajectories together" to "store only the trajectory of the final simulate call". Or alternatively, we could give `simulate` and optional name argument if you want to store multiple trajectories.

eb8680: Yes, re-initializing `self.trajectory` in `_pyro_simulate` seems more correct than the current behavior.

eb8680 (changes requested):
Can you add a test that checks the case where a `StaticInterruption`'s `time` coincides with a logging time, and also rerun the `dynamical_intro` notebook from scratch to verify that it no longer fails because of this bug?

SamWitty:
> Can you add a test that checks the case where a `StaticInterruption`'s `time` coincides with a logging time, and also rerun the `dynamical_intro` notebook from scratch to verify that it no longer fails because of this bug?

I believe this existing test covers the test case? https://github.com/BasisResearch/chirho/blob/sw-time-collision/tests/dynamical/test_log_trajectory.py#L42 Happy to add more if that doesn't cover what you're looking for.

I'll rerun the notebook in this PR.

Would you like me to make the change we discussed in this comment in this PR or separately?

> @eb8680 , do you think it makes sense to move this line into `_pyro_simulate` instead of `__enter__`. This would change the behavior from "concatenate multiple `simulate` calls trajectories together" to "store only the trajectory of the final simulate call". Or alternatively, we could give `simulate` an optional name argument if you want to store multiple trajectories.

SamWitty:
@eb8680 , I edited this PR to address my suggestion here: https://github.com/BasisResearch/chirho/pull/397#discussion_r1412110358 It is now ready for review again.

To ensure that handlers commute after this PR we now have the following strict order enforced regardless of the order handlers are applied:
`LogTrajectory` - Handles `simulate` during preprocessing to address potential time collision.
`Solver` - Handles `simulate` during postprocessing to do basically everything.
`BatchObservation` - Handles `simulate` to condition on data using a continuation, which occurs after all postprocessing.

I reran the dynamical system notebook, confirming that it works with these changes (although, requiring a minor change to the name of a trace address in the final cell because of a previous PR). I'm excluding the revised run from this PR to avoid merge conflicts with #377, which I expect to be completed by EOD or early tomorrow. Now that #308 is merged, I'll add a follow up PR after #377 to add the dynamical systems notebook to CI.

chirho/dynamical/handlers/interruption.py
```diff
@@ -130,5 +130,17 @@ def __init__(
         self.observation = observation
         super().__init__(times)
 
-    def _pyro_post_simulate(self, msg: dict) -> None:
-        self.trajectory = observe(self.trajectory, self.observation)
+    def _pyro_simulate(self, msg: dict) -> None:
```
eb8680, line 133: I don't think you should need to change `StaticBatchObservation` at all, other than adding a single `super()._pyro_post_simulate` call:
```
def _pyro_post_simulate(self, msg: dict) -> None:
    super()._pyro_post_simulate(msg)  # update self.trajectory
    self.trajectory = observe(self.trajectory, self.observation)

File: chirho/dynamical/handlers/trajectory.py [hunk 3]
```diff
@@ -23,9 +27,20 @@ def __init__(self, times: torch.Tensor):
...
+
         self.trajectory: State[T] = State()
-        return super().__enter__()
+
+        if start_time == self.times[0]:
+            # If we're starting at the beginning of the timespan, we need to log the initial state.
+            # LogTrajectory's simulate_point will log only timepoints that are greater than the start_time of each
+            # simulate_point call, which can occur multiple times in a single simulate call when there
+            # are interruptions.
+            self.trajectory = append(
```
eb8680, line 41: I think this change is right, and is already enough to make `LogTrajectory` and `Solver` commutative. In particular, since `append` is associative and `Solver` doesn't change `initial_state` or `start_time` in its `_pyro_simulate` method, as long as no outside code needs to access `self.trajectory` before the `simulate` call finishes, it shouldn't matter if `LogTrajectory`'s `_pyro_simulate` runs before or after `Solver`'s.

chirho/dynamical/internals/solver.py
```diff
@@ -22,6 +22,11 @@ class Solver(Generic[T], pyro.poutine.messenger.Messenger):
     @typing.final
     @staticmethod
     def _pyro_simulate(msg: dict) -> None:
+        msg["done"] = True
+
+    @typing.final
+    @staticmethod
+    def _pyro_post_simulate(msg: dict) -> None:
```
eb8680, line 29: This move from `_pyro_simulate` to `_pyro_post_simulate` seems unnecessary, per my comment on `LogTrajectory`.

eb8680 (commented):
Making sure `LogTrajectory`, `Solver` and `BatchObservation` commute is the right idea, but it seems like there's a simpler way to achieve that:

SamWitty:
> Making sure `LogTrajectory`, `Solver` and `BatchObservation` commute is the right idea, but it seems like there's a simpler way to achieve that:

I agree. The key challenge here is actually in the assignment of `self.trajectory: State[T] = State()` in `_pyro_simulate` of `LogTrajectory`. Even though the `append` is commutative, reseting `self.trajectory` must happen before the `Solver`' `_pyro_simulate` is called, or else it'll be entirely cleared in that line.

eb8680:
> The key challenge here is actually in the assignment of self.trajectory: State[T] = State() in _pyro_simulate of LogTrajectory. Even though the append is associative, reseting self.trajectory must happen before the Solver's _pyro_simulate is called, or else it'll be entirely cleared in that line.

Hmm, what about adding an extra boolean state variable to `LogTrajectory` that gets set in `_pyro_post_simulate` to track whether `simulate` has run? It's kind of clunky but at least it's isolated to `LogTrajectory`.
```python
def __init__(self, ...):
    self._needs_reset: bool = False
    self.trajectory = State()
    ...

def __enter__(self):
    self._needs_reset = False
    self.trajectory = State()
    ...

def _pyro_simulate(self, msg: dict) -> None:
    if self._needs_reset:
        self.trajectory = State()
        self._needs_reset = False
    ...

def _pyro_post_simulate(self, msg: dict) -> None:
    if not self._needs_reset:
        self._needs_reset = True
    ...
```

SamWitty:
> > The key challenge here is actually in the assignment of self.trajectory: State[T] = State() in _pyro_simulate of LogTrajectory. Even though the append is associative, reseting self.trajectory must happen before the Solver's _pyro_simulate is called, or else it'll be entirely cleared in that line.
> 
> Hmm, what about adding an extra boolean state variable to `LogTrajectory` that gets set in `_pyro_post_simulate` to track whether `simulate` has run? It's kind of clunky but at least it's isolated to `LogTrajectory`.
> 
> ```python
> def __init__(self, ...):
>     self._needs_reset: bool = False
>     self.trajectory = State()
>     ...
> 
> def __enter__(self):
>     self._needs_reset = False
>     self.trajectory = State()
>     ...
> 
> def _pyro_simulate(self, msg: dict) -> None:
>     if self._needs_reset:
>         self.trajectory = State()
>         self._needs_reset = False
>     ...
> 
> def _pyro_post_simulate(self, msg: dict) -> None:
>     if not self._needs_reset:
>         self._needs_reset = True
>     ...
> ```

My solution is in the spirit of this suggestion, but a bit less explicit and a bit more concise. I'm really happy either way.

PR #398, opened 2023-11-28
Staging branch for `chirho.robust` module development

SamWitty:


chirho/robust/internals/linearize.py
```diff
@@ -0,0 +1,158 @@
...
+        x += torch.where(not_converged, v * p, zeros_xr)
+        r -= torch.where(not_converged, v * z, zeros_xr)
+        newrdotr = torch.where(not_converged, torch.dot(r, r), newrdotr)
+        mu = torch.where(not_converged, newrdotr / rdotr, mu)
+        p = torch.where(not_converged, r + mu * p, p)
+        rdotr = torch.where(not_converged, newrdotr, rdotr)
+
+        # rdotr = newrdotr
+        # if rdotr < residual_tol:
+        #     break
```
agrawalraj, line 71: Maybe we should just delete this comment

chirho/robust/internals/linearize.py
```diff
@@ -0,0 +1,158 @@
...
+    if num_samples_inner is None:
+        num_samples_inner = num_samples_outer**2
+
+    predictive = pyro.infer.Predictive(
+        model,
+        guide=guide,
+        num_samples=num_samples_outer,
+        parallel=True,
+    )
+    predictive_params, func_predictive = make_functional_call(predictive)
```
agrawalraj, line 135: Why do we need to use `make_functional_call` here? `func_predictive` is used in L150, but we don't require gradients with respect to `data` anyways

chirho/robust/internals/linearize.py
```diff
@@ -0,0 +1,158 @@
...
+
+    log_prob = NMCLogPredictiveLikelihood(
+        model, guide, num_samples=num_samples_inner, max_plate_nesting=max_plate_nesting
+    )
+    log_prob_params, func_log_prob = make_functional_call(log_prob)
+    score_fn = torch.func.grad(func_log_prob)
+
```
agrawalraj, line 142: If `cg_iters` is not None, then we should make sure it's less than or equal to the dimension of guide parameters (`total_params`) because conjugate gradients gives us the exact inverse vector product in at most `total_params` iterations. So we could add something like this:

```
if cg_iters is not None:
    total_params = sum([param.numel() for param in log_prob_params.values()])
    # In at most total_params iterations, conjugate gradients yields exact inverse vector product 
    cg_iters = min(cg_iters, total_params)
```

chirho/robust/internals/linearize.py
```diff
@@ -0,0 +1,158 @@
...
+    cg_solver = functools.partial(
+        conjugate_gradient_solve, cg_iters=cg_iters, residual_tol=residual_tol
+    )
+
+    @functools.wraps(score_fn)
+    def _fn(point: Point[T], *args: P.args, **kwargs: P.kwargs) -> ParamDict:
```
agrawalraj, line 148: Given that we will usually like to evaluate `_fn` at multiple points, we might decorate this function with vmap:

```
@torch.vmap
@functools.wraps(score_fn)
def _fn(point: Point[T], *args: P.args, **kwargs: P.kwargs) -> ParamDict:
    ...
```

chirho/robust/internals/linearize.py
```diff
@@ -0,0 +1,158 @@
...
+        with torch.no_grad():
+            data: Point[T] = func_predictive(predictive_params, *args, **kwargs)
+            data = {k: data[k] for k in point.keys()}
+        fvp = make_empirical_fisher_vp(
+            func_log_prob, log_prob_params, data, *args, **kwargs
+        )
```
agrawalraj, line 154: I think these lines should be pulled out of `_fn` because (1) each time `_fn` is called a different approximation to the fisher information matrix is used since `data` is resampled, and (2) there's a performance hit because you need to resample `data` each time. This will be time consuming when `data` has a lot of Monte Carlo samples.

chirho/robust/ops.py
```diff
@@ -0,0 +1,48 @@
...
+        target = PredictiveFunctional(model, guide)
+    else:
+        target = functional(model, guide)
+
+    # TODO check that target_params == model_params | guide_params
+    assert isinstance(target, torch.nn.Module)
+    target_params, func_target = make_functional_call(target)
+
+    @functools.wraps(target)
+    def _fn(point: Point[T], *args: P.args, **kwargs: P.kwargs) -> S:
```
agrawalraj, line 42: If we change `linearized` to take in multiple points (see comment in `linearize.py`), then `_fn` should be a function of multiple points here as well.

chirho/robust/ops.py
```diff
@@ -0,0 +1,48 @@
...
+        param_eif = linearized(point, *args, **kwargs)
+        return torch.func.jvp(
+            lambda p: func_target(p, *args, **kwargs), (target_params,), (param_eif,)
+        )[1]
```
agrawalraj, line 46: If `param_eif` is batched over multiple points, then we should replace this line with something like this:

```
return torch.vmap(lambda d: torch.func.jvp(lambda p: func_target(p, *args, **kwargs), (target_params,), (d,))[1], in_dims=0)(param_eif)
```

agrawalraj (commented):
It's really coming together! Thanks for organizing everything so nicely :)

chirho/robust/internals/linearize.py
eb8680: I believe it's necessary for `vmap` to compose correctly with `linearize`.

chirho/robust/internals/linearize.py
eb8680: I don't think we can move these lines out of `_fn` because they depend on the values of `args` and `kwargs`, but I think we'd want a fresh approximation to the Fisher matrix every time anyway since `log_prob_params` could change (e.g. by a gradient update) even if `args` didn't.

chirho/robust/ops.py
eb8680: I think we can safely apply `vmap` outside of the single-point definitions of `linearize` and `influence_fn` to apply them to batches of points - `vmap`'s purpose is to enable exactly that separation of concerns.

chirho/robust/ops.py
eb8680: As discussed above, I don't think we need to worry about batching over points inside of `influence_fn` and `linearize`. We can just apply `vmap` to these functions in contexts where they are used.

chirho/robust/internals/linearize.py
agrawalraj: If we put *args and **kwargs as inputs to linearize, then we could pull out these lines, right? I think the odd behavior is that the resulting influence function is not consistent across points. For example, if I call `_fn` from linearize on the same point twice, you'll get two different answers. Similarly, it's harder to compare evaluations of `_fn` at different points because each evaluation uses a different approximation to the Fisher information matrix. 

I agree that if `log_prob_params` changes, then we would want a fresh approximation to the Fisher matrix.

eb8680:
Blocked by merge conflicts and #476

eb8680 (approved):
Will merge when tests pass

PR #400, opened 2023-11-29
Implement explain causes in a notebook

rfl-urbaniak:
In a notebook, illustrate how the existing explanation module can be used both to reconstruct the Halpern definition of explanation and to go beyond it by allowing the priors to impact the evaluation of potential explanations.


eb8680 (commented):
Needs to be updated to use `chirho.explainable`

rfl-urbaniak:
Closing,  superseded by #447.

PR #401, opened 2023-11-29
Explanation module, actual causality and explanation notebooks

rfl-urbaniak:
Once the explanation-related code is separated into a module and the actual causality and explanation notebooks have been completed and merged into `ru-causality-staging`,  this should land in `master`.


eb8680:
Closing since the staging branch we're using is `staging-causality` following #391

PR #404, opened 2023-11-29
Add upper bound on number of CG steps

eb8680:
Addresses @agrawalraj's comments in #398:
> If cg_iters is not None, then we should make sure it's less than or equal to the dimension of guide parameters (total_params) because conjugate gradients gives us the exact inverse vector product in at most total_params iterations.


agrawalraj (changes requested):
I think we don't want to do this here because then we can't vmap over `_fn` in `linearize`. I think we should move this check to `linearize`

eb8680:
@agrawalraj this is ready for review

agrawalraj (approved):
Looks good!

Issue #386, opened 2023-11-23
Fisher-vector product operation needs more tests


eb8680:
The core internal computation in `chirho.robust` is the product of the estimated Fisher matrix with a vector. Our implementation(s) of this operation should be extensively tested and profiled to improve reliability and performance.

Tests to add:
- [ ] Gold-standard cases where the Fisher matrix is known exactly
- [ ] Stronger tests of correct composition with batching and differentiation transforms from `torch.func`

PR #405, opened 2023-11-29
Added more tests for `linearize` and `make_empirical_fisher_vp`

agrawalraj:
Address #386 

- `robust_fixtures.py` contains utilities to remove repeated code across testing files (just as we did for the tests in `dynamical`)
- `test_linearize_against_analytic_ate` compares our empirical efficient influence function for average treatment effect against the analytical formula from Kennedy
- `test_empirical_fisher_vp_against_analytical` compares the empirical fisher vector product against the analytical fisher vector product for the special multivariate Gaussian case
- `test_fisher_vmap_smoke`  check composition with `torch.func.vmap`
- `test_fisher_grad_smoke`) check composition with `torch.func.grad`


agrawalraj:
> @agrawalraj , I removed the "status:awaiting review" tag because it looked like you were still working on this. Feel free to change it back if you want.

Thanks yeah I realized the tests weren't passing so made some small edits to get them to pass now.

agrawalraj:
> @agrawalraj I see #408 is marked as ready for review, does that mean this is ready as well?

The failing test here will be addressed by #408. Let me fix the linting error now.

tests/robust/test_internals_compositions.py
```diff
@@ -0,0 +1,59 @@
...
+)
+from chirho.robust.internals.predictive import NMCLogPredictiveLikelihood
+from chirho.robust.internals.utils import make_functional_call
+
+from .robust_fixtures import SimpleGuide, SimpleModel
+
+pyro.settings.set(module_local_params=True)
+
+
+def test_empirical_fisher_vp_nmclikelihood_cg_composition():
```
eb8680, line 19: If this is the test you expect to fail without #408, can you mark it with `pytest.mark.xfail(reason="fails without fix in https://github.com/BasisResearch/chirho/pull/408")`?

agrawalraj: Given that the change in #408 now just uses `SeedMessenger` in the `linearize` body, I can fix these errors by doing the same in these two tests. I'll push up these changes now.

tests/robust/test_internals_linearize.py
```diff
@@ -178,3 +174,173 @@ def test_nmc_param_influence_vmap_smoke(
...
+    empirical_fisher_vp_func({"loc": v_single})
+    empirical_fisher_vp_func_batched({"loc": v_batch})
+    try:
+        empirical_fisher_vp_func({"loc": v_batch})
+        assert False, "Fisher vector product should not work on a batch of vectors"
+    except RuntimeError:
+        pass
+    try:
+        empirical_fisher_vp_func_batched({"loc": v_single})
+        assert False, "Batched Fisher vector product should not work on a single vector"
```
eb8680, line 247: Can you use `with pytest.raises(RuntimeError)` contexts to check these failures instead of manual `try`/`except`?

tests/robust/test_internals_linearize.py
```diff
@@ -178,3 +174,173 @@ def test_nmc_param_influence_vmap_smoke(
...
+
+
+@pytest.mark.parametrize(
+    "data_config",
+    [
+        (torch.zeros(1, requires_grad=True), torch.eye(1)),
+        (torch.ones(2, requires_grad=True), torch.eye(2)),
+    ],
+)
+def test_fisher_vmap_smoke(data_config):
```
eb8680, line 224: I don't think this should just be a smoke test - with `vmap`, you can check that the `vmap`-ed output matches the output of evaluating the function on each item in the batch sequentially.

tests/robust/test_internals_linearize.py
```diff
@@ -178,3 +174,173 @@ def test_nmc_param_influence_vmap_smoke(
...
+    guide_train = pyro.infer.autoguide.AutoDelta(conditioned_model)
+    elbo = pyro.infer.Trace_ELBO()(conditioned_model, guide_train)
+
+    # initialize parameters
+    elbo(D_train)
+
+    adam = torch.optim.Adam(elbo.parameters(), lr=0.03)
+
+    # Do gradient steps
+    for _ in range(2000):
```
eb8680, line 317: How long does this test take to run? Can we do fewer than 2000 steps here?

tests/robust/robust_fixtures.py
```diff
@@ -0,0 +1,208 @@
...
+        for name, value in mle_est.items():
+            setattr(self, name + "_param", torch.nn.Parameter(value))
+
+    def forward(self, *args, **kwargs):
+        for name in self.names:
+            value = getattr(self, name + "_param")
+            pyro.sample(name, dist.Delta(value))
+
+
+def closed_form_ate_correction(X_test: Point[T], theta: ParamDict):
```
eb8680, line 196: nit: I think these types should be `typing.TypedDict`s since you're expecting specific names:
```
class ATETestPoint(TypedDict):
    X: torch.Tensor
    A: torch.Tensor
    Y: torch.Tensor


class ATEParamDict(TypedDict):
    propensity_weights: torch.Tensor
    outcome_weights: torch.Tensor
    treatment_weight: torch.Tensor
    intercept: torch.Tensor


def  closed_form_ate_correction(X_test: ATETestPoint, theta: ATEParamDict) -> Tuple[torch.Tensor, torch.Tensor]:
    ...
```

tests/robust/robust_fixtures.py
```diff
@@ -0,0 +1,208 @@
...
+class DataConditionedModel(PyroModule):
+    r"""
+    Helper class for conditioning on data.
+    """
+
+    def __init__(self, model: PyroModule):
+        super().__init__()
+        self.model = model
+
+    def forward(self, D: Point[T]):
```
eb8680, line 68: nit: type `D: Point[torch.Tensor]` since `DataConditionedModel` is not a `Generic` in `T`

tests/robust/test_internals_compositions.py
eb8680: That's fine, but remember to remove the `SeedMessenger` here in #408, since you want this test to exercise the changes made to `NMCLogPredictiveLikelihood` there.

tests/robust/robust_fixtures.py
```diff
@@ -0,0 +1,208 @@
...
+
+from chirho.observational.handlers import condition
+from chirho.robust.internals.utils import ParamDict
+from chirho.robust.ops import Point
+
+pyro.settings.set(module_local_params=True)
+T = TypeVar("T")
+
+
+class SimpleModel(PyroModule):
```
eb8680, line 17: An easy way to add a bit more test coverage would be to parameterize `SimpleModel` by a link function that returns a `Distribution` for `y` and add more test cases with different link functions.

tests/robust/test_internals_linearize.py
agrawalraj: Done.

tests/robust/test_internals_linearize.py
agrawalraj: It currently takes 13.3seconds. I reduced this down to 500 steps now but this isn't the bottleneck. The main bottleneck is computing the EIF since we need a large number of Monte Carlo samples (here 10k) to approximate the EIF well.

tests/robust/robust_fixtures.py
agrawalraj: Got it thanks!

tests/robust/robust_fixtures.py
agrawalraj: Done

tests/robust/test_internals_compositions.py
agrawalraj: Ok in #408, I followed your suggestion here https://github.com/BasisResearch/chirho/pull/408#pullrequestreview-1760679539, and did not actually modify `NMCLogPredictiveLikelihood`. I just checked if 

```
log_prob = NMCLogPredictiveLikelihood(
        model, guide, num_samples=3, max_plate_nesting=3
    )
log_prob_params, func_log_prob = make_functional_call(log_prob)
func_log_prob = SeedMessenger(123)(func_log_prob)
```
works in the `test_nmc_likelihood_seeded` test in #408. I think we can move this discussion to #408 if you have other suggestions on this point!

eb8680 (approved):
LGTM

tests/robust/test_internals_linearize.py
eb8680: OK, that's fine, we can revisit if it becomes a bottleneck.

Issue #407, opened 2023-12-01
Issue composing `make_empirical_fisher_vp`, conjugate gradients, and `NMCLogPredictiveLikelihood`


agrawalraj:
In the tests so far in #405, I've been focusing on the MLE case (i.e., when the guide is just equal to a point estimate) since I have closed-form analytical expressions for testing purposes. I've been trying to construct more tests for more general guides, and I think there's an issue composing `make_empirical_fisher_vp`, conjugate gradients, and `NMCLogPredictiveLikelihood` in general. There are several issues all basically rooted in the fact that each call to `NMCLogPredictiveLikelihood` is random, namely if `log_prob = NMCLogPredictiveLikelihood(model, guide)` and `call_one = log_prob(x_n)` and `call_two =  log_prob(x_n)`, then `call_one != call_two` unless the guide is just a point estimate guide.

**Issue 1**:
Our empirical fisher info equals

1/N \sum_{n=1}^N grad log_prob(x_n)  grad log_prob(x_n)^T. Since `log_prob(x_n) !=  log_prob(x_n)` for two successive calls, the empirical Fisher matrix is no longer a symmetric matrix. In general, vanilla conjugate gradients do not work for non-symmetric matrices but there are more general CG methods. For a counterexample, see https://math.stackexchange.com/questions/4795473/conjugate-gradient-descent-when-a-is-non-singular-but-not-symmetric. 

**Issue 2:**
Conjugate gradients assumes that there is a *fixed* matrix A, and we want to solve for x in Ax = b. In our implementation, there is not a fixed A across CG iterations because `fvp(v) != fvp(v)`, where `fvp = make_empirical_fisher_vp(func_log_prob, log_prob_params, data)`. Again this occurs because `log_prob` is randomized.

**Potential solution**:
One option would be to sample the particles in `NMCLogPredictiveLikelihood` in `__init__` so that successive calls would be deterministic. That would fix "Issue 2". Assuming we make this change, there are two options for addressing "Issue 1". 

_Option 1_: Use 1/N \sum_{n=1}^N grad log_prob(x_n) grad log_prob(x_n)^T, where log_prob has particles sampled in `__init__` instead of `forward` method. Then, this matrix is symmetric but biased because E[grad log_prob(x_n) grad log_prob(x_n)^T] != E[grad log_prob(x_n)] E[grad log_prob(x_n)]^T since the **same** particles are used for both calls to log_prob.

[...]

PR #408, opened 2023-12-01
Make `NMCLogPredictiveLikelihood` seeded

agrawalraj:
The current implementation of `NMCLogPredictiveLikelihood` has non-deterministic behavior which creates issues with computing fisher vector products. This PR addresses this non-deterministic behavior by seeding calls and adds the test `test_nmc_likelihood_seeded` to make sure this seeding composes with `vmap`, `jvp` and `vjp`. 

See also the discussion in #407.


eb8680 (commented):
Using PRNG state to pin the samples is a good idea, but rather than using the lower-level primitive `set_rng_seed` directly and baking this behavior into the implementation of `NMCLogPredictiveLikelihood`, you could make this change more modular and more local by applying the higher-level `pyro.poutine.SeedMessenger` to the result of `make_functional_call(NMCLogPredictiveLikelihood(...))` within the body of `linearize`.

The issue you were seeing before with `SeedMessenger` was caused by the unfortunate fact that Pyro handlers applied to `nn.Module`s return ordinary functions, not another `nn.Module` with the parameters still exposed.

eb8680:
Since this change is independent of #405, can you change the base branch to `staging-robust` so that the PR targets #398?

agrawalraj:
> Since this change is independent of #405, can you change the base branch to `staging-robust` so that the PR targets #398?

Since I branched off `ra-sw-fisher-tests` here to reuse some of the testing utility functions, we might need to merge in #405 first before this PR now.

agrawalraj:
> Using PRNG state to pin the samples is a good idea, but rather than using the lower-level primitive `set_rng_seed` directly and baking this behavior into the implementation of `NMCLogPredictiveLikelihood`, you could make this change more modular and more local by applying the higher-level `pyro.poutine.SeedMessenger` to the result of `make_functional_call(NMCLogPredictiveLikelihood(...))` within the body of `linearize`.
> 
> The issue you were seeing before with `SeedMessenger` was caused by the unfortunate fact that Pyro handlers applied to `nn.Module`s return ordinary functions, not another `nn.Module` with the parameters still exposed.

Sorry I somehow missed this comment. Will address now!

agrawalraj:
@eb8680 I added another check to `test_nmc_likelihood_seeded` to check that all calls to `func_log_prob` are the same. Something very odd happens where the first call is different but all subsequent calls are the same (see the failing assert on L78 in `test_nmc_likelihood_seeded`). Have you noticed a similar issue when using `SeedMessenger`?

agrawalraj:
> Do you get the same error if you provide a value for `max_plate_nesting`?

Got it thanks! Yes the error goes away when I set `max_plate_nesting` value.

agrawalraj:
> > Do you get the same error if you provide a value for `max_plate_nesting`?
> 
> Got it thanks! Yes the error goes away when I set a value for `max_plate_nesting`.

Since the issue stems from `max_plate_nesting`, do you think we should make `max_plate_nesting` a required argument to `NMCLogPredictiveLikelihood` and `PredictiveFunctional` @eb8680?

eb8680:
> do you think we should make max_plate_nesting a required argument to NMCLogPredictiveLikelihood and PredictiveFunctional instead of setting it's value in the forward method

I'd prefer not to make it required if we can avoid it. You can set it by hand in tests, and we can try and extract it automatically earlier in `linearize` before that logic is triggered within `NMCLogPredictiveLikelihood`.

agrawalraj:
> > do you think we should make max_plate_nesting a required argument to NMCLogPredictiveLikelihood and PredictiveFunctional instead of setting it's value in the forward method
> 
> I'd prefer not to make it required if we can avoid it. You can set it by hand in tests, and we can try and extract it automatically earlier in `linearize` before that logic is triggered within `NMCLogPredictiveLikelihood`.

Cool that makes sense to me

chirho/robust/internals/linearize.py
```diff
@@ -138,6 +140,7 @@ def linearize(
         model, guide, num_samples=num_samples_inner, max_plate_nesting=max_plate_nesting
     )
     log_prob_params, func_log_prob = make_functional_call(log_prob)
+    func_log_prob = SeedMessenger(seed)(func_log_prob)
```
eb8680, line 143: This will fix the seed at the level of `linearize`, so that `linearize` and `influence_fn` return deterministic functions even though they are computing (unbiased) random quantities. Is that really what we want?

I thought what was necessary to address #407 was only that the seed to stay fixed for the duration of a single `cg_solver` call, which would mean changing the output `_fn` of `linearize` like so:
```python
def _fn(point: Point[T], *args: P.args, **kwargs: P.kwargs) -> ParamDict:
    ...
    with SeedMessenger(get_rng_seed()):
        return cg_solver(fvp, point_score)

return _fn
```

eb8680: How about using a helper function to pin the rng state across `fvp` calls in `cg_solver`?
```python
@contextlib.contextmanager
def reset_rng_state(rng_state):
    try:
        prev_rng_state = get_rng_state()
        set_rng_state(rng_state)
    finally:
        set_rng_state(rng_state)

...

def linearize(...):
    ...
    def _fn(point: Point[T], *args, **kwargs):
        ...
        pinned_fvp = reset_rng_state(get_rng_state())
        return cg_solver(pinned_fvp, point_score)

    return _fn
```

agrawalraj: Cool thanks let me give this a try!

chirho/robust/internals/linearize.py
```diff
@@ -90,7 +91,7 @@ def make_empirical_fisher_vp(
     batched_func_log_prob: Callable[[ParamDict, Point[T]], torch.Tensor] = torch.vmap(
         lambda p, data: func_log_prob(p, data, *args, **kwargs),
         in_dims=(None, 0),
-        randomness="different",
+        randomness="same",
```
eb8680, line 94: Could you explain this change? IIUC this will reuse randomness within `NMCLogPredictiveLikelihood` across test data samples in a batch, which will affect the statistical properties of our Fisher matrix estimator.

agrawalraj: I made this change because we no longer need randomness to change with the Hessian formulation of Fisher in #428. In #428, we can approximate Fisher with

$$
-\sum_{n=1}^N \nabla^2_{\phi} \log \sum_{t=1}^T p(x_n \mid \theta_t), \ \theta_t \sim q(\theta \mid \phi).
$$ 

Since there's only one call to $p(x_n \mid \theta_t)$, we can share randomness across datapoints unlike in the score formulation of Fisher. However, I guess it doesn't really hurt to have randomness even in the Hessian formulation case. So I'll move it back to "different".

eb8680 (approved):
Looks good, thanks for fixing this!

PR #431, opened 2023-12-07
Backend Implementation with DifferentialEquations.jl — Does not Include Interruptions

azane:
Adds dynamical systems solver backend relying on DifferentialEquations.jl.

Differences/nuances wrt the torchdiffeq backend.

1. For gradients to work, dynamics must be a pure function of both state and parameters. The current design choice around this is that parameters must be included in the initial state, and the dynamics should return a dstate comprising a set of keys that is a subset of those in the initial state dictionary — the returned keys are assumed to be state variables, and those that are not returned by the dynamics, but are included in the initial state, are assumed to be parameters.
2. Dynamics are defined in python and are functions of `State[np.ndarray]`, even though the trajectories returned, and initial state are all to be `torch.Tensor`. This nuance is due to the fact that `juliacall` interoperates with numpy arrays when calling and taking return values from python functions. The dynamics are executed by julia from within julia, meaning the juliacall nuances apply here.
3. Complexity arises in that more complex julia types (e.g. dual numbers, symbolics, etc.) are, by default, very poorly transparent to the python-defined dynamics (see [here](https://github.com/SciML/juliatorch/issues/14#issue-2017683001) for more details). Primarily, numpy cannot `asarray` these julia types because it (presumably?) (incorrectly?) introspects the julia entity's memory buffer, and mistakenly allocates arrays with >32 dimensions — exceeding the numpy limit. Additionally, numpy ufuncs (e.g. np.exp) do not work on julia values, and julia functions (e.g. jl.exp) do not work on numpy arrays. After a few hours of research, it seemed that no off-the-shelf numpy solution (including specifying a dtype of object) can prevent this. As such, the solution here wraps the julia entities in python objects that 1) forward dunder methods on to the underlying julia entity and 2) intercept numpy ufunc calls and reroutes them to the corresponding julia calls. This is almost all done dynamically. I hope there are better/cleaner ways to do this (perhaps from the julia side), but it works for now.
4. It's assumed that the dynamics function will be jit compileable. An effectful `_lazily_compile_solver` is handled by the `DiffEqDotJl` solver, such that a compiled SciML `ODEProlem` is built upon first needing it. If a user tries to use that same solver instance a second time with different dynamics, then it raises an error.

To the end of keeping the PR non-huge, this does not include an implementation of interruptions for this backend. Preliminary investigations suggest this will be feasible, though could necessitate some adjustments to the interface shared with torchdiffeq. Luckily, the interruption callbacks needn't be compiled along with the problem — this should simplify things considerably.

blocking #432


azane:
Note that clean-up, organization, and additional comments/documentation is still pending, but otherwise ready for review.

azane:
Also need to make sure new dependencies are correctly installed for tests to run.

chirho/dynamical/internals/backends/diffeqdotjl.py
```diff
@@ -0,0 +1,558 @@
...
+    #  user's dynamics only have to handle numpy arrays, and not also torch tensors. This is fine, as the only way
+    #  we use the initial_dstate below is for its keys, which comprise the state variables.
+    initial_state_ao_params_np = {k: copy(v.detach().numpy()) for k, v in initial_state_ao_params.items()}
+    # TODO unify this time business with how torchdiffeq is doing it?
+    if 't' in initial_state_ao_params_np:
+        raise ValueError("Initial state cannot contain a time variable 't'. This is added on the backend.")
+    initial_state_ao_params_np['t'] = t0.detach().numpy()
+
+    # Run the dynamics on the converted initial state.
+    initial_dstate_np = dynamics(initial_state_ao_params_np)
```
azane, line 439: cc @SamWitty 

Do a check here to return a more informative message saying that dynamics for diffeqdotjl backend have to support numpy arrays

chirho/dynamical/internals/backends/diffeqdotjl.py
```diff
@@ -0,0 +1,558 @@
...
+from juliatorch import JuliaFunction
+import juliacall
+import numpy as np
+from typing import Union
+from functools import singledispatch
+from copy import copy
+from juliacall import Main as jl
+
+
+class _DunderedJuliaThingWrapper:
```
azane, line 21: Put this stuff in another file and make backends/diffeqdotjl a package.

azane:
Superseded by https://github.com/BasisResearch/chirho_diffeqpy

PR #432, opened 2023-12-07
Refactor of Tests to Evaluate DiffEqDotJl Backend — Tests Involving Interruptions Excluded

azane:
Parameterizes tests in `test_log_trajectory` and `test_static_observations` — the two files primarily testing features not involving interruptions, which haven't been implemented for the backend yet — to test multiple backends with the same tests.

This has become messy primarily due the fact that the julia backend requires pure dynamics, while torchdiffeq can support impure dynamics (referencing parameters not appearing in the arguments of the dynamics function). The tests were designed for the latter, and don't interoperate particularly well with the former. Am still working through how best to refactor the tests to clean this up.

Regardless, tests do pass for `LogTrajectory` and point observations that don't involve interruptions.

merge blocked by #431 

Perhaps approval of #431 should probably be conditional on these tests passing? That said, tests of basic functionality are included in this PR.


azane:
Superseded by https://github.com/BasisResearch/chirho/pull/554

PR #433, opened 2023-12-07
Make State into a type alias of Mapping

eb8680:
Addresses #350 

This refactoring PR makes the interface type `chirho.dynamical.ops.State` into an alias for the generic protocol `typing.Mapping[str, T]`, rather than a concrete `dict` type.


eb8680:
@SamWitty I think after this lands we should cut a `v0.2.0` release. The other issues in https://github.com/BasisResearch/chirho/milestone/9 don't need to block that.

PR #436, opened 2023-12-07
Fix header formatting in dynamical_intro

eb8680:
One of the outline entries in `dynamical_intro.ipynb` is rendered incorrectly by Sphinx.


SamWitty:
I also just noticed that the lists (e.g. S(t)) are also being renderred incorrectly.

Issue #337, opened 2023-10-17
Missing documentation for `chirho.dynamical`


eb8680:
The following submodules of `chirho.dynamical` are currently missing documentation:
- [x]  `chirho.dynamical.ops`
- [x] `chirho.dynamical.handlers.interruptions`
- [x] `chirho.dynamical.handlers.trajectory`
- [x] `chirho.dynamical.handlers.solver`

eb8680:
This was resolved by #439

PR #439, opened 2023-12-07
Add documentation for `dynamical.ops` and `dynamical.handlers`

azane:
Partly addresses #337 by adding documentation for `simulate`, `LogTrajectory` `StaticBatchObservation` and all user-facing interruptions.


docs/source/type_aliases.rst
azane, line 1: Not sure if we actually want this, but this was the cleanest way I could find (after looking only a little bit) to document these type aliases. Happy to remove if not desired.

SamWitty:
@azane , could you address the lint failure? Thanks!

SamWitty:
Confirming that operations and handlers render correctly when I build locally:

Screenshot 2023-12-08 at 10 57 11 AM

However, as my comment above about the type aliases warning implies, I don't think the type aliases are currently included in the docs build.

SamWitty:
> One thing I notice in the above rendering is that the `State` type alias has been evaluated away. E.g. rather than `State[T]` it shows up as `Mapping[str, T]`. This is maybe fine, but it will make things confusing if we use the type alias name directly in the documentation itself. The example I showed above has `State[T]` listed as the return type.

@azane and I figured out a workaround, but will plan on submitting a follow-up
PR as this issue is pervasive throughout all of the documentation, not just dynamical systems.

PR #440, opened 2023-12-08
Add new `SimpleModel` and `SimpleGuide`

### LINKED ISSUE #427: `SimpleModel` and `SimpleGuide` used for testing has a singular Fisher information matrix
In many of the tests in `robust`, we have been using the following model and guide:

```python
class SimpleModel(PyroModule):
    def forward(self):
        a = pyro.sample("a", dist.Normal(0, 1))
        with pyro.plate("data", 3, dim=-1):
            b = pyro.sample("b", dist.Normal(a, 1))
            return pyro.sample("y", dist.Normal(a + b, 1))


class SimpleGuide(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.loc_a = torch.nn.Parameter(torch.rand(()))
        self.loc_b = torch.nn.Parameter(torch.rand((3,)))

    def forward(self):
        a = pyro.sample("a", dist.Normal(self.loc_a, 1))
        with pyro.plate("data", 3, dim=-1):
            b = pyro.sample("b", dist.Normal(self.loc_b, 1))
            return {"a": a, "b": b}
```

I analytically worked out the Fisher information matrix of the guide parameters, and the result is

$$ I(\text{loc}_a, \text{loc}_b) = 
	\begin{bmatrix} 
	1 & 1 & 0 & 0 \\
	1 & 0 & 1 & 0 \\
	1 & 0 & 0 & 1 \\
	\end{bmatrix}^T
	\begin{bmatrix} 
	3 & 1 & 1 \\
	1 & 3 & 1 \\
	1 & 1 & 3 \\
	\end{bmatrix}
	\begin{bmatrix} 
	1 & 1 & 0 & 0 \\
	1 & 0 & 1 & 0 \\
	1 & 0 & 0 & 1 \\
	\end{bmatrix}
	=
\begin{bmatrix} 
15 & 5 & 5 & 5 \\
5 & 3 & 1 & 1 \\
5 & 1 & 3 & 1 \\
5 & 1 & 1 & 3 \\
\end{bmatrix}
$$

Since the middle matrix is a 3x3 matrix, the maximum rank of the 4x4 Fisher matrix is 3, and is therefore non-invertible. This lack of invertibility makes conjugate gradients unstable in the test here: https://github.com/BasisResearch/chirho/blob/0148b615aece1484f616dc5d7d1c2301e9efdf0f/tests/robust/test_internals_compositions.py#L20.

More fundamentally, the reason why the Fisher information is non-invertible is because `SimpleModel` is not identifiable. See also Theorem on page 18 here https://www.icts.res.in/sites/default/files/INFOASI-2017-01-02-Lecture-Mark-Transtrum-Lecture1A.pdf which states how identifiability is closely linked with the invertibility of Fisher information. 

Given these issues, I think we should create a better template model and guide for testing purposes.

agrawalraj:
Closes #427


tests/robust/robust_fixtures.py
```diff
@@ -15,11 +15,18 @@
 
 
 class SimpleModel(PyroModule):
+    def __init__(
+        self,
+        link_fn: Callable[..., dist.Distribution] = lambda mu: dist.Normal(mu, 1.0),
```
SamWitty, line 20: Is this used anywhere?

PR #441, opened 2023-12-08
staging-causality (explainable reasoning as a separate module)

rfl-urbaniak:
This is the staging scene for the explainable reasoning module.


eb8680 (approved):
I think this is stable enough now to land on `master`.

PR #446, opened 2023-12-08
add revised actual causality notebook

rfl-urbaniak:
Refactoring of the explainable reasoning module required small changes to imports and code chunks in the notebook. The notebook has also been equipped with a smoke test, and added to `test_notebooks.sh`.


rfl-urbaniak:
Seems like in python 3.10 the test fails due to unrelated reasons,  `test_cut_module_discrete` fails.

eb8680 (approved):
Looks good, thanks for adding the example locations.

PR #447, opened 2023-12-08
revise the causal explanation notebook, added test

rfl-urbaniak:
Updating the notebook in light of the refactoring.


eb8680 (commented):
Can you explain the role of the `condition_on_consequent` argument to `ExplanationEvaluation` in the notebook, and why it's set to `False` throughout? I thought you always wanted to condition on the consequents in the factual world, since otherwise there's nothing to explain.

rfl-urbaniak:
Added the following explanation:

In principle, we want to be able to condition on the observed consequents: as whatever happened to the consequents is what we want to explain. On the other hand, some of the conditions in the original definition involve queries in the unconditioned model (e.g., *are the consequents different always if we intervene on a given candidate cause?*) In a more general setting to mimic the definition from the book we would need to have two separate model runs, with and without conditioning and run inference to propagate the results of the conditioning. In this context, when we're dealing with not too many discrete variables, we can proceed in a simplified manner and get away with running forward with an unconditioned model. The conditioning on the consequent is then delegated to trace handling for those parts of the definition that require conditioning. We will emphasize this element of trace handling conditioning in a comment when it happens.

eb8680:
Let's land #441 first. I'm not entirely convinced by that reasoning about conditioning on consequents, but this notebook doesn't need to block #441.

eb8680:
@rfl-urbaniak can you resolve these new merge conflicts?

SamWitty:
@rfl-urbaniak and @eb8680 , what is the current status of this PR? I have some spare cycles if you'd like my help reviewing.

rfl-urbaniak:
Closing as superseded by later developments.

Issue #463, opened 2023-12-20
Optimization when guide is a point estimate in `chirho.robust`


agrawalraj:
See item 1 in #459.

eb8680:
Resolved by #473

Issue #429, opened 2023-12-06
Extract `max_plate_nesting` in `linearize` call to prevent seeding issue with  `NMCLogPredictiveLikelihood`


agrawalraj:
See https://github.com/BasisResearch/chirho/pull/408#issuecomment-1843684611

eb8680:
Resolved by #473

PR #473, opened 2023-12-30
Replace some `torch.vmap` usage with a hand-vectorized `BatchedNMCLogPredictiveLikelihood`

eb8680:
Addresses #463 (the remaining issue 1 from #459)
Also addresses #429

Naive usage of `torch.vmap` in `chirho.robust` has caused a number of performance issues. This PR refactors the contents of `chirho.robust.internals.predictive` to replace `NMCLogPredictiveLikelihood` with a custom `BatchedNMCLogPredictiveLikelihood` that makes use of the Pyro-specialized batching logic from `chirho.indexed`.

Remaining tasks:
- [x] Address `vmap` regression/bug causing tests to fail under `torch>=2.1.0` that passed under `torch==2.0.1`
- [x] Verify that #463 is indeed resolved by these changes
- [x] Add unit tests for new components in `chirho.robust.internals.utils`

### DIFF
--- a/chirho/observational/handlers/condition.py
+++ b/chirho/observational/handlers/condition.py
@@ -1,4 +1,4 @@
-from typing import Callable, Generic, Hashable, Mapping, TypeVar, Union
+from typing import Callable, Generic, Mapping, TypeVar, Union
 
 import pyro
 import torch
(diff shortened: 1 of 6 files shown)


agrawalraj (approved):
Looks really great to me, thanks!

PR #474, opened 2024-01-03
Reproduce DR learner notebook using `chirho.robust` 

### LINKED ISSUE #373: Redo DR-learner example using our new interface
~Blocked by #372~

We should be able to reproduce the results in https://basisresearch.github.io/chirho/dr_learner.html using our automated efficient influence function approximated using Monte Carlo.

agrawalraj:
This PR closes #373


eb8680:
Needs merge conflict resolution following merge of #398 to `master`

docs/source/automated_dr_learner.ipynb
SamWitty, line 1: I believe the outline is taken from the backdoor adjustment tutorial, and doesn't align with this notebook.

SamWitty (changes requested):
This notebook could certainly use some more text, but I think we can add that in a subsequent PR. 

See my comment about the outline for an easy to address issue that should be fixed before the PR lands.

docs/source/automated_dr_learner.ipynb
eb8680: I think the outlines are wrong or at least have broken links in a number of other tutorials, which suggests the manually written outlines may be too brittle to bother maintaining when Sphinx, JupyterLab, VSCode and most other editors already generate and render outlines automatically from the section headers. I just removed it from this one entirely.

SamWitty: Sounds good. We can follow up in a separate PR to make all the outlines consistent with whatever we decide.

SamWitty (approved):

Issue #468, opened 2023-12-22
Implement targeted maximum likelihood estimator in `chirho.robust.handlers.estimators`


agrawalraj:
See `tmle` in https://www.notion.so/basisresearch/Design-document-robust-statistics-with-influence-functions-0a3ecaab5b944e2fade867a6958039b7

SamWitty:
Notes from conversation with Raj:

It would be great if we could find an example where the **root solving** itself also has an analytic solution.

agrawalraj:
Two relevant papers that describe TMLE estimator: https://www.degruyter.com/document/doi/10.2202/1557-4679.1043/html?lang=en and https://arxiv.org/pdf/2306.08598.pdf and

PR #484, opened 2024-01-07
Implement TMLE estimator using Influence Functions

SamWitty:
This pull request implements a first-pass version of TMLE using our implementation of approximate influence functions. This will need to be revised to abstract out (1) the approximation method for the efficient influence function and (2) the optimization method(s) used to solve the maximum likelihood search over \epsilon and the projection to guide parameters.

Addresses #468


eb8680:
Needs merge conflict resolution before it can land on `master` following #398

eb8680:
Needs merge conflict resolution to target `master` following #398

PR #487, opened 2024-01-08
Make functional argument to influence_fn required

eb8680:
This refactoring PR removes the default behavior for the `functional` argument of `influence_fn`, and makes a value for this argument mandatory.


SamWitty (approved):
lgtm

PR #491, opened 2024-01-10
allow for empty witness in SearchForCause

rfl-urbaniak:
allows `witnesses = {}` and adds a clause to the test of `SearchForCause`.


eb8680 (commented):
Can you add the same fix for `antecedents` and `consequents`? There are checks below this logic that `antecedents` and `consquents` are nonempty but the same bug will prevent them from being reached.

rfl-urbaniak:
will do

rfl-urbaniak:
by the way, the difference is though the user might intend the witness set to be empty (this is the but-for clause check), but if they're not specifying the antecedent or consequent, this should perhaps raise a value error. Something to think about in the future. For now I'll just do the same thing.

eb8680 (approved):

Issue #480, opened 2024-01-05
Support higher-order influence computation


eb8680:
Our implementation of `chirho.robust.ops.influence_fn` estimates the first-order efficient influence function, but there are some settings where computing higher-order influence functions may enable additional bias correction (e.g. https://arxiv.org/abs/0805.3040 or https://arxiv.org/abs/2305.04116 ).

As with ordinary derivatives, the discussion in ["Higher Order Tangent Spaces and Influence Functions"](https://arxiv.org/abs/1502.00812) suggests it should be possible to compute higher-order influence functions and correction terms by applying a first-order influence estimator to itself recursively:
![image](https://github.com/BasisResearch/chirho/assets/2032320/881bc0b4-9c21-41bb-9c44-472205305079)

This recursion should not be difficult to implement using our existing machinery, although nesting our naive Monte Carlo gradient estimators may not be computationally feasible beyond order 4 or so.

Issue #479, opened 2024-01-05
Support influence functions for n-ary functionals


eb8680:
The current implementation of `chirho.robust.ops.influence_fn` is defined only for functionals of one distribution, but there are many functionals of interest in machine learning, e.g. the KL divergence or mutual information, that are defined over 2 or more distributions and are known to benefit from multivariate versions of one-step correction: https://proceedings.neurips.cc/paper_files/paper/2015/hash/06138bc5af6023646ede0e1f7c1eac75-Abstract.html

Extending our implementation of `influence_fn` to these cases is conceptually straightforward - we just compute a tuple containing the influence of each individual distribution with the others held fixed.

Here is code illustrating the binary case in terms of the existing unary `influence_fn`, with guides suppressed as in #477:
```python
def influence2(
    functional: Callable[[Callable[P, Any], Callable[P, Any]], Callable[P, S]],
    model1: Callable[P, Any],
    model2: Callable[P, Any],
    **linearize_kwargs
) -> Callable[P, Tuple[S, S]]:
    def _inf(point1: Point[A], point2: Point[B], *args, **kwargs) -> Tuple[S, S]:
        return (
            influence_fn(model1, lambda m: functional(m, model2))(point1, *args, **kwargs),
            influence_fn(model2, lambda m: functional(model1, m))(point2, *args, **kwargs),
        )
    return _inf
```

PR #492, opened 2024-01-10
Make influence_fn a higher-order Functional

eb8680:
Addresses #479, #480 

This PR changes the type signature of `chirho.robust.ops.influence_fn` to make it more compositional, following the suggestions in #479 and #480. In particular, `influence_fn` is now a higher-order functional that maps a target functional to another functional:
```py
def influence_fn(functional: Functional[P, S], *points: Point[T]) -> Functional[P, S]: ...
```
It also generalizes the `Functional` interface (and therefore those of `influence_fn` and `linearize`) to arbitrary arities (but still only implements the unary case):
```py
class Functional(Protocol[P, S]):
    def __call__(self, __model: Callable[P, Any], *models: Callable[P, Any]) -> Callable[P, S]: ...
```
Bundling these interface changes together in one PR will hopefully ensure that there will only be one set of merge conflicts associated with #479 and #480 to resolve in other open PRs targeting `staging-robust`.

### DIFF
--- a/chirho/robust/handlers/estimators.py
+++ b/chirho/robust/handlers/estimators.py
@@ -1,6 +1,6 @@
-from typing import Any, Callable, TypeVar
+from typing import TypeVar
 
-from typing_extensions import Concatenate, ParamSpec
+from typing_extensions import ParamSpec
 
 from chirho.robust.ops import Functional, Point, influence_fn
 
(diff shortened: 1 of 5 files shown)


agrawalraj (approved):
lgtm!

PR #493, opened 2024-01-11
Implement `InferConstraints`

### LINKED ISSUE #448: Provide handler for extracting random variable supports
In the `SearchForExplanation` handler added in #445, the interface requires a significant amount of information to be specified about a model up front.

To make this easier, we should add a handler `InferSupports` to `chirho.explainable.handlers.alternatives` that grabs the name and support of each random variable. We could try to implement this in terms of `pyro.poutine.trace`, which already has the necessary information, or define it directly as below:
```python
class InferSupports(pyro.poutine.messenger.Messenger):
    supports: Mapping[str, pyro.distributions.constraints.Constraint]

    def _pyro_post_sample(self, msg: dict) -> None:
        if not pyro.poutine.util.site_is_subsample(msg):
            self.supports[msg["name"]] = msg["fn"].support
```
`InferSupports` can then be used to construct arguments for `SearchForExplanation`:
```python
with InferSupports() as s:
    model(...)

antecedents: Mapping[str, Observation] = ...
antecedent_alternatives: Mapping[str,  = {a: support for a, support in s.support.items() if a in antecedents}

with SearchForExplanation(antecedent_alternatives=antecedent_alternatives, ...):
    ...
```

rfl-urbaniak:
Resolves #448 , introduces the `InferConstraints` messenger,  a Pyro Messenger for inferring distribution constraints.

    :param antecedents: (optional) A list or mapping of antecedent variables or constraints.
    :param witnesses: (optional) A list or mapping of witness variables or constraints.
    :param consequents: (optional) A list or mapping of consequent variables or constraints.

    :return: An instance of InferSupports with new attributes: ``supports``, ``antecedents``,
        ``witnesses``, and ``consequents``. ``supports`` is a dictionary mapping variable names
        to constraints for all variables in the model, while the other attributes are analogous,
        restricted to variables mentioned in the optional parameters (empty dictionaries if no
        optional parameters are provided).

    Example:

        >>> def mixed_supports_model():
        >>>     uniform_var = pyro.sample("uniform_var", dist.Uniform(1, 10))
        >>>     normal_var = pyro.sample("normal_var", dist.Normal(3, 15))
        >>> with InferSupports(antecedents=["uniform_var"]) as s:
        ...      mixed_supports_model()
        >>> print(s.supports, s.antecedents)


chirho/explainable/constraints_experiment.ipynb
eb8680, line 1: Can you remove this notebook? If there's code there you'd like to keep, you can move that code to the unit tests or to the docstring of `InferSupports`.

chirho/explainable/internals/defaults.py
```diff
@@ -173,3 +174,90 @@ def _soft_neq_independent(support: constraints.independent, v1: T, v2: T, **kwar
     for _ in range(support.reinterpreted_batch_ndims):
         result = torch.sum(result, dim=-1)
     return result
+
+
+class InferSupports(pyro.poutine.messenger.Messenger):
```
eb8680, line 179: Can you move this to `handlers/components.py` so that it's available to user code and rename it to `ExtractSupports`? I know I suggested `InferSupports` in #448 but I think `ExtractSupports` is more accurate since the information from this handler is only available after it exits.

chirho/explainable/internals/defaults.py
```diff
@@ -173,3 +174,90 @@ def _soft_neq_independent(support: constraints.independent, v1: T, v2: T, **kwar
...
+        >>> with InferSupports(antecedents=["uniform_var"]) as s:
+        ...      mixed_supports_model()
+        >>> print(s.supports, s.antecedents)
+    """
+
+    supports: MutableMapping[str, pyro.distributions.constraints.Constraint]
+
+    def __init__(
+        self,
+        antecedents: Optional[
```
eb8680, line 207: I don't think these arguments are necessary. This handler isn't specific to usage with `SearchForExplanation`, so the distinction between antecedents, witnesses and consequents should just be enforced by code that accesses the results of this handler in its `self.supports`.

chirho/explainable/internals/defaults.py
```diff
@@ -173,3 +174,90 @@ def _soft_neq_independent(support: constraints.independent, v1: T, v2: T, **kwar
...
+                List[str],
+                Mapping[str, Intervention[T]],
+                Mapping[str, constraints.Constraint],
+            ]
+        ] = None,
+    ):
+        super(InferSupports, self).__init__()
+
+        for group in ["antecedents", "witnesses", "consequents"]:
+            setattr(self, group, self._extract_keys(locals()[group]))
```
eb8680, line 232: As with the arguments to `__init__`, I think these attributes and the information they represent should be kept separate from this handler.

chirho/explainable/internals/defaults.py
```diff
@@ -173,3 +174,90 @@ def _soft_neq_independent(support: constraints.independent, v1: T, v2: T, **kwar
...
+        elif isinstance(input, dict):
+            return list(input.keys())
+        else:
+            return list(input)  # overspecified to pass lint
+
+    def _pyro_post_sample(self, msg: dict) -> None:
+        if not pyro.poutine.util.site_is_subsample(msg):
+            self.supports[msg["name"]] = msg["fn"].support  # t ype: ignore
+
+    def __exit__(self, exc_type, exc_value, traceback):
```
eb8680, line 251: If you remove the group attributes, you should be able to delete this method entirely.

chirho/explainable/internals/defaults.py
```diff
@@ -173,3 +174,90 @@ def _soft_neq_independent(support: constraints.independent, v1: T, v2: T, **kwar
...
+        ] = None,
+    ):
+        super(InferSupports, self).__init__()
+
+        for group in ["antecedents", "witnesses", "consequents"]:
+            setattr(self, group, self._extract_keys(locals()[group]))
+
+        self.supports = {}
+
+    def _extract_keys(
```
eb8680, line 236: If you remove the group attributes, you should be able to delete this method entirely.

eb8680 (changes requested):
This is a great start, but I think you can simplify this quite a bit by removing the grouping information from the handler.

chirho/explainable/internals/defaults.py
rfl-urbaniak: done

chirho/explainable/internals/defaults.py
rfl-urbaniak: removed

chirho/explainable/internals/defaults.py
rfl-urbaniak: removed

chirho/explainable/internals/defaults.py
rfl-urbaniak: Yes

chirho/explainable/constraints_experiment.ipynb
rfl-urbaniak: Sure

eb8680 (changes requested):
Looks like you forgot to revert your previous changes to `internals/defaults.py` and `test_internals.py`

eb8680 (approved):

Issue #496, opened 2024-01-12
Soft conditioning functions should live in observational module


eb8680:
In #472 we added helpers `soft_eq` and `soft_neq` to the `chirho.explainable` module. After the staging branch PR #441 is merged, we should move these functions to `chirho.observational.handlers.soft_conditioning` and reimplement the existing functionality there in terms of `soft_eq`.

Tasks:
- [x] Move to observational module #500
- [x] Remove old soft conditioning code #534

PR #500, opened 2024-01-12
migrate soft conditioning and fix a few small explainable module documentation typos

rfl-urbaniak:
Addresses #496 . In the process, noticed inconsistency in variable names rendering in the documentation and an outdated piece containing a factual error, fixed this as well.


eb8680 (approved):
This is a good start, thanks. In a followup PR, can you remove the old TorchKernel and RBFKernel classes defined in soft_conditioning.py and simplify AutoSoftConditioningReparam to use soft_eq directly?

PR #508, opened 2024-01-18
Finite Difference Baseline

azane:
Adds finite difference machinery in a way that can be used in experiments.


docs/examples/robust_paper/finite_difference_eif/distributions.py
```diff
@@ -0,0 +1,34 @@
+from .abstractions import ModelWithMarginalDensity, FDModelFunctionalDensity
+from scipy.stats import multivariate_normal
+import pyro
+import pyro.distributions as dist
+
+
+class MultivariateNormalwDensity(ModelWithMarginalDensity):
```
agrawalraj, line 7: In a separate PR, could you add this to `MODEL` in `statics.py` as an element of `MultivariateNormalModel`.

docs/examples/robust_paper/utils.py
```diff
@@ -112,3 +113,13 @@ def get_mle_params_and_guide(conditioned_model, n_iters=2000, lr=0.03):
         k: v.clone().detach().requires_grad_(True) for k, v in guide_train().items()
     }
     return theta_hat, MLEGuide(theta_hat)
+
+
+@contextmanager
```
agrawalraj, line 118: Can you just use `reset_rng_state` from `chirho.robust.internals.utils` here instead?

agrawalraj (changes requested):
LGTM, just a few small changes!

agrawalraj (approved):
lgtm

docs/examples/robust_paper/finite_difference_eif/distributions.py
azane: https://github.com/BasisResearch/chirho/pull/510/commits/719cff8d8d7d3da9932e9673593c3c23bdf5f14d

PR #509, opened 2024-01-19
Refactor of `influence_approx.py`

agrawalraj:
This now includes ATE functional and pointwise Monte Carlo EIF

@azane you can slot your stuff in in `influence_approx.py` L142.


azane (approved):
maybe pull analytic and chirho method into functions so the experiment function is easier to read

PR #517, opened 2024-02-13
Fix memory leak issue in `linearize`

### LINKED ISSUE #516: Memory leak when calling `linearize` multiple times
After running 100 iterations, 40GB of memory is used. After 250 iterations, memory runs out.

```python
import torch
import pyro
import pyro.distributions as dist
from pyro.infer import Predictive

from chirho.robust.internals.utils import ParamDict
from chirho.robust.internals.linearize import linearize
from chirho.robust.handlers.predictive import PredictiveModel

pyro.settings.set(module_local_params=True)


class ToyNormal(pyro.nn.PyroModule):
    def forward(self):
        mu = pyro.sample("mu", dist.Normal(0.0, 1.0))
        sd = pyro.sample("sd", dist.HalfNormal(1.0))
        return pyro.sample(
            "Y",
            dist.Normal(mu, scale=sd),
        )


class ToyNormalKnownSD(pyro.nn.PyroModule):
    def __init__(self, sd_true):
        super().__init__()
        self.sd_true = sd_true

    def forward(self):
        mu = pyro.sample("mu", dist.Normal(0.0, 1.0))
        sd = pyro.sample("sd", dist.HalfNormal(1.0))
        return pyro.sample(
            "Y",
            dist.Normal(mu, scale=self.sd_true),
        )


class GroundTruthToyNormal(pyro.nn.PyroModule):
    def __init__(self, mu_true, sd_true):
        super().__init__()
        self.mu_true = mu_true
        self.sd_true = sd_true

    def forward(self):
        return pyro.sample(
            "Y",
            dist.Normal(self.mu_true, scale=self.sd_true),
        )


class MLEGuide(torch.nn.Module):
    def __init__(self, mle_est: ParamDict):
        super().__init__()
        self.names = list(mle_est.keys())
        for name, value in mle_est.items():
            setattr(self, name + "_param", torch.nn.Parameter(value))

    def forward(self, *args, **kwargs):
        for name in self.names:
            value = getattr(self, name + "_param")
            pyro.sample(
                name, pyro.distributions.Delta(value, event_dim=len(value.shape))
            )


N_pts = 500
mu_true = 0.0
sd_true = 1.0
true_model = GroundTruthToyNormal(mu_true, sd_true)
D_pts = Predictive(true_model, num_samples=N_pts, return_sites=["Y"])()
Y_pts = D_pts["Y"]
Y_pts = torch.sort(Y_pts).values


for i in range(100):
    print(i)
    theta_true = {
        "mu": torch.tensor(mu_true, requires_grad=True),
        "sd": torch.tensor(sd_true, requires_grad=True),
    }
    model = ToyNormal()
    guide = MLEGuide(theta_true)

    # Linearize model
    monte_eif = linearize(
        PredictiveModel(model, guide), num_samples_outer=10000, num_samples_inner=1
    )({"Y": Y_pts})
```

agrawalraj:
This PR adds an argument `detach` to `linearize` and `make_empirical_fisher_vp` to avoid an odd memory leak when gradients are not detached in `make_empirical_fisher_vp`. In case people would like to autodiff through `linearize` (e.g., for higher-order influence function computations), then `detach` should be set to `False`. 

Resolves #516.

### DIFF
--- a/chirho/robust/ops.py
+++ b/chirho/robust/ops.py
@@ -16,8 +16,7 @@ Point = Mapping[str, Observation[T]]
 class Functional(Protocol[P, S]):
     def __call__(
         self, __model: Callable[P, Any], *models: Callable[P, Any]
-    ) -> Callable[P, S]:
-        ...
+    ) -> Callable[P, S]: ...
 
 
 def influence_fn(
@@ -127,7 +126,7 @@ def influence_fn(
         if len(models) != len(points):
             raise ValueError("mismatch between number of models and points")
 
-        linearized = linearize(*models, **linearize_kwargs)
+        linearized = linearize(*models, **linearize_kwargs)  # type: ignore
         target = functional(*models)
 
         # TODO check that target_params == model_params
--- a/chirho/robust/internals/linearize.py
+++ b/chirho/robust/internals/linearize.py
@@ -112,6 +112,7 @@ def make_empirical_fisher_vp(
     batched_func_log_prob: Callable[Concatenate[ParamDict, Point[T], P], torch.Tensor],
     log_prob_params: ParamDict,
     data: Point[T],
+    detach: bool = True,
     *args: P.args,
     **kwargs: P.kwargs,
 ) -> Callable[[ParamDict], ParamDict]:
@@ -213,7 +214,11 @@ def make_empirical_fisher_vp(
 
         # Perlmutter's trick
         vjp_fn = torch.func.vjp(jvp_fn, log_prob_params)[1]
-        return vjp_fn(-1 * mean_vector)[0]  # Fisher = -E[Hessian]
+        val = vjp_fn(-1 * mean_vector)[0]  # Fisher = -E[Hessian]
+        # Detach to memory leak if we do not need to backpropagate
+        # through the Fisher vector product:
+        # https://github.com/BasisResearch/chirho/issues/516
+        return {k: v.detach() if detach else v for k, v in val.items()}
 
     return _empirical_fisher_vp
 
@@ -226,6 +231,7 @@ def linearize(
     cg_iters: Optional[int] = None,
     residual_tol: float = 1e-4,
     pointwise_influence: bool = True,
+    detach: bool = True,
 ) -> Callable[Concatenate[Point[T], P], ParamDict]:
     r"""
     Returns the influence function associated with the parameters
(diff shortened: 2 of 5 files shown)


SamWitty:
@agrawalraj , I'm not seeing a test that confirms that this addresses the memory leak issue. Could you please add a test that shows: (i) when using `detach=False` the memory footprint increases for two successive calls to `influence_fn` and (ii) when using `detach=True` the memory footprint does not increase for two successive calls?

chirho/robust/ops.py
```diff
@@ -127,7 +126,7 @@ def _influence_functional(*models: Callable[P, Any]) -> Callable[P, S]:
         if len(models) != len(points):
             raise ValueError("mismatch between number of models and points")
 
-        linearized = linearize(*models, **linearize_kwargs)
+        linearized = linearize(*models, **linearize_kwargs)  # type: ignore
```
SamWitty, line 129: Why is this needed?

SamWitty (changes requested):

chirho/robust/ops.py
agrawalraj: This is unrelated / can remove this line. I added it because the linter complains that linearized isn't typed.

agrawalraj:
@SamWitty I added `test_memory_leak.py` to track how memory changes when `detach` is `True` or `False`. I didn't make it a formal test because this issue might be specific to `torch.func` and could be resolved in a future release of `torch.func`. The memory leak when `detach = False` is weird because I'm not saving any variables in the loop...

SamWitty:
> @SamWitty I added `test_memory_leak.py` to track how memory changes when `detach` is `True` or `False`. I didn't make it a formal test because this issue might be specific to `torch.func` and could be resolved in a future release of `torch.func`. The memory leak when `detach = False` is weird because I'm not saving any variables in the loop...

Fair enough. I think a good approach then is to just add a version with `detach = True` to the tests, making sure that it doesn't accumulate a memory footprint. This way we'll be able to confirm that this actually addresses the memory bug, while having the tests be robust to upstream changes in `torch.func`.

eb8680:
Can this be addressed using the [`torch.no_grad()`](https://pytorch.org/docs/stable/generated/torch.no_grad.html#no-grad) context manager without adding a flag to the API?

agrawalraj:
> Can this be addressed using the [`torch.no_grad()`](https://pytorch.org/docs/stable/generated/torch.no_grad.html#no-grad) context manager without adding a flag to the API?

I tried adding this flag and that works! Should we just close this PR then? My only concern though would be people might not know to use `torch.no_grad()` and then might run into this memory issue.

agrawalraj:
> > @SamWitty I added `test_memory_leak.py` to track how memory changes when `detach` is `True` or `False`. I didn't make it a formal test because this issue might be specific to `torch.func` and could be resolved in a future release of `torch.func`. The memory leak when `detach = False` is weird because I'm not saving any variables in the loop...
> 
> Fair enough. I think a good approach then is to just add a version with `detach = True` to the tests, making sure that it doesn't accumulate a memory footprint. This way we'll be able to confirm that this actually addresses the memory bug, while having the tests be robust to upstream changes in `torch.func`.

Good idea. I changed the file to explicitly test with `detach=True`. However, we might abandon this PR b/c you can avoid this memory issue w/ `torch.no_grad()`

agrawalraj:
Closing this PR in favor of #519.

PR #522, opened 2024-02-20
Bump Pyro to `>1.9.0` and fix resulting lint errors

SamWitty:
TODO


SamWitty:
@eb8680 , I started this thinking it would be easy, but it appears to be very much not easy to address these linting errors. A major source of trouble is our use of generic types throughout ChiRho, and with `pyro==1.9.0` the results of primitive pyro operations are now fixed as `torch.Tensor` or related.

eb8680:
Addressed in #527

Issue #524, opened 2024-02-21
Initial states don't broadcast properly in `chiro.dynamical`


rfl-urbaniak:
Initial states do not broadcast as expected, which leads to shape errors when using `Predictive` with `parallel=True`, or when
using both `pyro.plate` and `MultiWorldCounterfactual`. A minimal failing example:

```
class UnifiedFixtureDynamics(pyro.nn.PyroModule):
    def __init__(self, beta=None, gamma=None):
        super().__init__()

        self.beta = beta
        self.gamma = gamma

    def forward(self, X: State[torch.Tensor]):
        dX: State[torch.Tensor] = dict()

        dX["S"] = -self.beta * X["S"] * X["I"]
        dX["I"] = self.beta * X["S"] * X["I"] - self.gamma * X["I"]  # noqa
        dX["R"] = self.gamma * X["I"]
        return dX

init_state = dict(S=torch.tensor(10.0), I=torch.tensor(3.0),
                          R=torch.tensor(3.0))

start_time = torch.tensor(0.0)
end_time = torch.tensor(10.0)
logging_times = torch.linspace(start_time + 0.01, end_time - 2, 5)

def bayesian_sir():
        with TorchDiffEq():
                    with MultiWorldCounterfactual() as cf:
                        beta = pyro.sample("beta", dist.Beta(18, 600))
                        gamma = pyro.sample("gamma", dist.Beta(1600, 1600))
                        fixture = UnifiedFixtureDynamics(beta = beta, gamma = beta)
                        cf_state = simulate(
                            fixture,
                            init_state,
                            start_time,
                            end_time,
                        )

# fails, but will not fail if num_samples = 1
predictive = Predictive(bayesian_sir, num_samples=2, parallel=True)
sir_samples = predictive()
```

A workaround, discovered together with @SamWitty , is to introduce the init state using `pyro.sample` statements, which ensures proper init state broadcasting, as follows:

```
def bayesian_sir():
    with LogTrajectory(
       times=logging_times, is_traced = True
    ) as dt:
        with TorchDiffEq():
            with StaticIntervention(time=intervene_time, intervention=intervene_state):
                with StaticIntervention(
                    time=intervene_time + 0.5, intervention=intervene_state
                ):
                    with MultiWorldCounterfactual() as cf:
                        beta = pyro.sample("beta", dist.Beta(18, 600))
                        gamma = pyro.sample("gamma", dist.Beta(1600, 1600))

[...]

PR #525, opened 2024-02-22
Fix TorchDiffEq tensor broadcasting

SamWitty:
This small PR addresses #524 by calling the partially evaluated `deriv` method a single time to determine any tensor broadcasting that occurred inside of its body. We then take the resulting tensor shapes and use them to define the broadcasted tensor shapes of the input state.

### DIFF
--- a/chirho/dynamical/internals/backends/torchdiffeq.py
+++ b/chirho/dynamical/internals/backends/torchdiffeq.py
@@ -105,8 +105,11 @@ def _batched_odeint(
     # TODO support event_dim > 0
     event_dim = 0  # assume states are batches of values of rank event_dim
 
+    # Call deriv once to get the shape of the output.
+    dy0 = func(t[..., 0], y0)
+
     y0_batch_shape = torch.broadcast_shapes(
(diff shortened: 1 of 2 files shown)


SamWitty:
@eb8680 , see the state of the branch and the failing tests at [5960b4c](https://github.com/BasisResearch/chirho/pull/525/commits/5960b4c48b7ea1e47515b901a7489569da0a2e90) for confirmation that this PR addresses the problem.

rfl-urbaniak:
Thanks @Sam! Tested on my side, this seems to solve the `init_state` broadcasting issues with the dynamics I've been dealing with, but does not resolve broadcasting issues in general. The derivative broadcasting issue still persists and the `torch.tensor([0.0]) * X["I"]` workaround is still needed for parallel sampling to work with `SIRDynamicsLockdown`.

SamWitty:
@rfl-urbaniak , thanks for pointing that out. I believe that error is a result of the model not being written in a way that is broadcastable, as the output is a tensor with shape 1 regardless of the input shape of `X["l"]`. Instead, we should always write these models with the following instead.

`dX["l"] = torch.zeros_like(X["l"])`

eb8680 (approved):

PR #527, opened 2024-02-23
Update type annotations for Pyro 1.9

eb8680:
Supercedes #522


eb8680:
This is basically done, but it would take more work to support older Pyro versions alongside Pyro 1.9.0 and I'm not sure that's worth the effort. However, if we do start to require `pyro-ppl>=1.9.0`, then this PR and the associated ChiRho release should wait until next week or the week after to avoid breaking ASKEM hackathon code.

eb8680:
This should not be merged before the ASKEM evaluation ends in 2 weeks. It also seems likely that there will be a Pyro 1.9.1 release soon that will address some of the issues here, so maybe we should wait for that as well.

SamWitty:
@eb8680 , now that the ASKEM evaluation is over would you like me to review this PR?

eb8680:
If possible I think we should aim to make this backward compatible with Pyro 1.8 before merging. That way we can unpin the Pyro version here in ChiRho and separately pin it to `<1.9.0` downstream in PyCIEMSS.

As I mentioned in a comment above, I had tried to enable backward compatibility when I made this PR, but it ended up being more work than I anticipated because there are backwards-incompatible changes to some types in Pyro 1.9.0. Unless I can come up with a simpler workaround for this problem, I'm not sure I'll have the time to fix it myself and land this PR in the next couple of weeks.

SamWitty (approved):
I have a bunch of miscellaneous comments here, none of which _really_ need to be addressed before this merged. That said, if I'm correct and the types for `msg` throughout could be specified as `Mapping[str, Any]` then that would be nice.

I'm approving, but I would appreciate some brief discussion on the points I brought up.

eb8680:
Closing as obsolete / addressed by #586

PR #528, opened 2024-02-23
Reuse non-default names across multiple splits in MultiWorldCounterfactual

eb8680:
This PR changes the behavior of the `MultiWorldCounterfactual` handler. Previously, `MultiWorldCounterfactual` created a fresh index variable for every `split()` call, regardless of the `name` keyword argument passed to `split`. After this PR, `MultiWorldCounterfactual` only introduces fresh variables for fresh (or empty) names.

This change is necessary to support a more sophisticated counterfactual semantics in dynamical systems.

### DIFF
--- a/chirho/counterfactual/handlers/counterfactual.py
+++ b/chirho/counterfactual/handlers/counterfactual.py
@@ -143,10 +143,14 @@ class MultiWorldCounterfactual(IndexPlatesMessenger, BaseCounterfactualMessenger
 
     @classmethod
     def _pyro_split(cls, msg: Dict[str, Any]) -> None:
-        name = msg["name"] if msg["name"] is not None else cls.default_name
-        index_plates = get_index_plates()
-        if name in index_plates:
-            name = f"{name}__dup_{len(index_plates)}"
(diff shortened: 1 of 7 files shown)


SamWitty:
@eb8680 , once this lands I'd like to add a tiny PR that adds an optional name kwarg to the existing dynamical systems interventions, which should the counterfactual behavior from "one world per state variable" to "one world per static/dynamic intervention". That will also hopefully make it much easier to index into worlds produced from dynamical systems interventions. I know this PR is already intended to lead to interventions on dynamics, but figured it would be worth mentioning how it will improve the ergonomics (with a minor change) of the existing dynamical systems interventions. Thoughts?

eb8680:
I guess this is ready for review, but since it's a breaking change I would be hesitant to merge it before the ASKEM evaluation ends in 2 weeks.

SamWitty:
@eb8680 , now that the ASKEM evaluation is over can I review this?

eb8680:
Sure, go for it.

SamWitty (approved):
lgtm

PR #529, opened 2024-02-27
explainable reasoning with SIR 

rfl-urbaniak:
This is the first example of a working application of the explainable module to a dynamical system. In a context in which multiple policies interact (here,  lockdown weakens the impact of a masking policy) and can in principle result in similar outcomes (overdetermination), the explainable module can be used to identify what actually caused the outcome. In this example, the outcome is overshoot (rate of S at the peak of I - final rate of S), which was chosen as it is a non-linear function of the params in an unintervened model, and is intuitively not straightforward to predict for intervened models (for instance, the direction in which it changes also becomes dependent on intervention times), while still being an outcome of interest in policy-making.


eb8680:
@rfl-urbaniak @PoorvaGarg can this PR be closed in favor of #565?

rfl-urbaniak:
> @rfl-urbaniak @PoorvaGarg can this PR be closed in favor of #565?

I think #565 is a continuation of this work on SIR, if @PoorvaGarg confirms, then yes.

eb8680:
Included in #565

PR #531, opened 2024-03-02
changed `lockdown_end` to sample from the correct uniform distribution

djinnome:
Before fixing `lockdown_end`:

image

After fixing `lockdown_end`

Screen Shot 2024-03-01 at 9 51 36 PM


SamWitty (approved):
good catch, thanks!

PR #478, opened 2024-01-04
Separate `influence_fn` operations from estimator semantics using effect handlers

SamWitty:
This refactoring PR separates out the `influence_fn` operation, which can be used in a way that is approximation-agnostic, from the effect handlers that implement their approximation. 

In particular, this extracts what was previously included in the `influence_fn` operation and ports that functionality into a new `MonteCarloInfluenceEstimator` effect handler.

In addition, this PR modifies the tests and some documentation to reflect these changes.

blocking: #482

PR #532, opened 2024-03-05
Separate influence function approximation from `influence_fn` syntax using handlers

SamWitty:
This refactoring PR separates out the influence_fn operation, which can be used in a way that is approximation-agnostic, from the effect handlers that implement their approximation.

In particular, this extracts what was previously included in the influence_fn operation and ports that functionality into a new MonteCarloInfluenceEstimator effect handler.

In addition, this PR modifies the tests and some documentation to reflect these changes.

Subsumes #478.

### DIFF
--- a/chirho/robust/handlers/estimators.py
+++ b/chirho/robust/handlers/estimators.py
@@ -1,13 +1,95 @@
-from typing import Any, Callable, TypeVar
+import warnings
+from typing import Any, Callable
 
+import pyro
 import torch
-from typing_extensions import ParamSpec
 
-from chirho.robust.ops import Functional, Point, influence_fn
+from chirho.robust.ops import Functional, P, Point, S, T, influence_fn
 
-P = ParamSpec("P")
-S = TypeVar("S")
-T = TypeVar("T")
+
+class MonteCarloInfluenceEstimator(pyro.poutine.messenger.Messenger):
+    """
+    Effect handler for approximating efficient influence functions with nested monte carlo.
+    See the MC-EIF estimator in https://arxiv.org/pdf/2403.00158.pdf for details and
+    :func:`~chirho.robust.ops.influence_fn` for example usage.
+
+    .. note::
+
+        * ``functional`` must compose with ``torch.func.jvp``
+        * Since the efficient influence function is approximated using Monte Carlo, the result
+          of this function is stochastic, i.e., evaluating this function on the same ``points``
+          can result in different values. To reduce variance, increase ``num_samples_outer`` and
+          ``num_samples_inner`` in ``linearize_kwargs``.
+        * Currently, ``model`` cannot contain any ``pyro.param`` statements.
+          This issue will be addressed in a future release:
+          https://github.com/BasisResearch/chirho/issues/393.
+        * There are memory leaks when calling this function multiple times due to ``torch.func``.
+          See issue:
+          https://github.com/BasisResearch/chirho/issues/516.
+          To avoid this issue, use ``torch.no_grad()`` as shown in the example above.
+
+    """
+
+    def __init__(self, **linearize_kwargs):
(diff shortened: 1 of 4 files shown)


SamWitty:
@eb8680 , this is ready for review except for linting errors that I'm having trouble addressing on my own. Could you take a look at the lint CI and let me know if you can think of a quick fix?

chirho/robust/handlers/estimators.py
```diff
@@ -1,13 +1,95 @@
...
+
+    """
+
+    def __init__(self, **linearize_kwargs):
+        self.linearize_kwargs = linearize_kwargs
+        super().__init__()
+
+    def _pyro_influence_fn(self, msg) -> None:
+        functional: Functional[P, S] = msg["args"][0]
+        points: Point[T] = msg["args"][1:]
```
eb8680, line 39: As the type checker error says, these type variables (`P`, `S` and `T`) are unbound. You can only use type variables in variable annotations if the type variables are introduced in an enclosing generic function type signature or `typing.Generic` type, so you need to either move the definition of `functional` and `points` inside `_influence_functional` (whose signature binds `P` and `S`), add `Generic[P, T, S]` as a parent class, or remove the variable annotations.

eb8680 (commented):
Because `influence_fn` is a higher-order function, there is some ambiguity in how its meaning should be modified by effect handlers. The design in this PR treats the higher-order function as an effectful operation.

However, all of the actual effectful computation happens when the function returned by `influence_fn` is called, not when `influence_fn` itself is called. That means that the meaning of `influence_fn` is determined by two separate handler contexts (the context in which `influence_fn` is called and the context in which the result of `influence_fn` is called).

To avoid that extra complexity, I would suggest making `influence_fn` a pure function that returns an effectful operation, rather than an effectful operation in its own right.

chirho/robust/handlers/estimators.py
SamWitty: Makes sense. Addressing your suggestion below simplified this substantially.

SamWitty:
@eb8680 , I took your suggestion to have `influence_fn` return an effectful operation, rather than have `influence_fn` itself be effectful. This PR is now ready for review.

eb8680 (commented):
This version still doesn't satisfy the desideratum that only a single handler context should affect the computation in `influence_fn`. All of the computation happens when `influence_fn(functional, point)(model)(*model_args)` is called, so the effectful operation handled by `MonteCarloInfluenceEstimator` should be the value
```py
model_influence: Callable[P, S]
model_influence = influence_fn(functional, point)(model)
```
not the intermediate values
```py
influence_fn: Callable[[Functional, Point], Functional]
```
or
```py
influence_functional: Functional
influence_functional = influence_fn(functional, point)
```
Otherwise, there are tricky semantic leakages:
```py
def f(model):
    ...
    with OtherInfluenceEstimator():
        return influence_fn(...)(model)
...
g = f(model)
...
with MonteCarloInfluenceEstimator():
    g()
```
In this example, as of this PR the meaning of the call `g()` is affected by the hidden `OtherInfluenceEstimator()` context within the body of `f`, not the more local `MonteCarloInfluenceEstimator()` context later on.

SamWitty:
@eb8680 , I made a pass at the changes you requested. I think it's at least close now. Some things I'd like your feedback on:

1. Would it better to define an effectful operation `influence` outside of the definition of `influence_fn` and `influence_functional` and then just call it?
2. Now the `linearize` computation in the `MonteCarloInfluenceEstimator` is computed lazily during execution of `influence`, whereas before it was pre-computed in the `influence_functional` layer. Does this matter? I'm not sure how else  to apply the handler only at the `influence` call otherwise.

eb8680 (approved):

PR #534, opened 2024-04-04
simplify `AutoSoftConditioningReparam` to use `soft_eq` directly.

rfl-urbaniak:
This is a follow-up to PR #500 .

The strategy:

1. Given a constraint and a scale/alpha passed as arguments to  `AutoSoftConditioning` define ` _soft_eq` that uses these values, and pass it on to `KernelSoftConditionReparam`, with proper constraint diagnosis preceding this call.

2. Eliminate `TorchKernel`, `SoftEqKernel` and `RBFKernel` completely.

3. Simplify `AutoSoftConditioning` accordingly.

Remark. I'm not sure if:

```
scale = self.scale * functools.reduce(
                 operator.mul, msg["fn"].event_shape, 1.0
             )
```

is still needed, commented out for now.

4. Modify `tests/observational/test_handlers.py` accordingly, to avoid using these kernels.


rfl-urbaniak:
This is sufficient for `tests/observational/test_handlers.py` to pass, but `tests/dynamical/test_handler_composition.py` fails.

1. Reinstated 

```
scale = scale * functools.reduce(
               operator.mul, msg["fn"].event_shape, 1.0
            )
```
still, same failure.

2.   Considered reverting to the intermediate variant that preserves `TorchKernel`  as a `torch.nn.Module`, same failure.

@SamWitty let's chat about this at some point.

chirho/observational/handlers/soft_conditioning.py
```diff
@@ -274,12 +225,8 @@ def configure(
             return None
 
         if msg["fn"].base_dist.v.is_floating_point():
-            scale = self.scale * functools.reduce(
-                operator.mul, msg["fn"].event_shape, 1.0
-            )
-            return KernelSoftConditionReparam(
-                RBFKernel(scale=scale, event_dim=len(msg["fn"].event_shape))
-            )
+            support = constraints.real
```
eb8680, line 228: You'll need to account for event dimensions here to make tests pass. I think you can just set `support = msg["fn"].base_dist.support`, which should be correct by construction.

chirho/observational/handlers/soft_conditioning.py
```diff
@@ -288,12 +235,15 @@ def configure(
             torch.int32,
             torch.int64,
         ):
-            alpha = self.alpha * functools.reduce(
-                operator.mul, msg["fn"].event_shape, 1.0
-            )
-            return KernelSoftConditionReparam(
-                SoftEqKernel(alpha=alpha, event_dim=len(msg["fn"].event_shape))
-            )
+            support = constraints.boolean
```
eb8680, line 238: As above, you'll need to account for event dimensions here. I think you can just delete this conditional and set `support = msg["fn"].base_dist.support`.

chirho/observational/handlers/soft_conditioning.py
```diff
@@ -254,7 +205,7 @@ class AutoSoftConditioning(pyro.infer.reparam.strategies.Strategy):
         the site's ``event_dim`` and ``support``.
     """
 
-    def __init__(self, *, scale: float = 1.0, alpha: float = 1.0):
+    def __init__(self, *, scale: float = 1.0, alpha: float = 1e-8):
```
eb8680, line 208: I think you can delete the `alpha` keyword argument from `AutoSoftConditioning`, since `scale` is the only parameter used by `soft_eq` and we're delegating the semantics of soft conditioning entirely to `soft_eq`.

chirho/observational/handlers/soft_conditioning.py
```diff
@@ -288,12 +235,15 @@ def configure(
...
-                SoftEqKernel(alpha=alpha, event_dim=len(msg["fn"].event_shape))
-            )
+            support = constraints.boolean
+            scale = self.alpha
+
+        def _soft_eq(v1: torch.Tensor, v2: torch.Tensor) -> torch.Tensor:
+            return soft_eq(support, v1, v2, scale=scale)
+
+        # TODO decide if still needed
+        scale = scale * functools.reduce(operator.mul, msg["fn"].event_shape, 1.0)
```
eb8680, line 245: I think you can remove this rescaling step. I'm actually not even sure I should have added it in the first place, but regardless I think `AutoSoftConditioning` should no longer do anything other than set up calls to `soft_eq`. If we do want logic like that, it can live somewhere inside `soft_eq`.

chirho/observational/handlers/soft_conditioning.py
rfl-urbaniak: Sure, will do - but keep in mind this means that even with a boolean constraints, `AutoSoftConditioning` will use the default scale of 1, which doesn't result in intuitively useful shape of values for this constraint.

eb8680 (approved):

PR #535, opened 2024-04-04
introduce notebook cleaning and linting scripts with an example

### LINKED ISSUE #409: Notebooks should be linted and typechecked in CI
It would be nice to have a linting and typechecking stage in CI for notebooks to go along with the smoke test stage added in #283.

Tasks:
- [ ] Add options and scripts to `make lint-notebooks` and `make format-notebooks` to our testing `Makefile`
- [ ] Fix all notebooks so that `make lint-notebooks` passes locally
- [ ] Add another CI workflow that calls `make lint-notebooks`

rfl-urbaniak:
This partially resolves #409 .

- two scripts for cleaning and linting notebooks are introduced
- `Makefile` is modified accordingly
- scripts are used on `tutorial_i` only for now. To pass lint, all imports are moved to the top of the file, and two function names are modified. Other changes are the result of auto-formatting. 
- `nbqa` is added to dependencies.

What is not covered in this PR:
- implementing this for other notebooks 
- introducing a corresponding workflow
reason: we need to make sure we agree about these methods first

Once we are on the same page about this, I will add the workflow and work on other notebooks one by one, to cover the same notebooks as the ones tested in `test_notebooks`. This will most likely require manual revisions to notebooks and re-running them completely, so I prefer to do these in separate small PRs.


SamWitty:
This looks good to me. One thing we might want to consider is just having `make format` and `make lint` run both regular code and notebook versions of the shell scripts rather than adding a new make command specific to notebooks.

@eb8680 , any thoughts?

eb8680:
> To pass lint, all imports are moved to the top of the file, and two function names are modified. Other changes are the result of auto-formatting.

I haven't used `nbqa` before. Just so I understand, what part of the imports wasn't fixed by autoformatting? Does this mean running `nbqa isort --check <filename>` can still fail even if you run `nbqa isort <filename>` first?

SamWitty:
> Relatedly, I resolved this in the collab repo by adding autoflake --remove-all-unused-imports --in-place ... to the cleaning script. I assume we do not want to remove unused imports in modules as, for example, this is a bit annoying for modules that are WIP, but perhaps could be useful for notebook cleaning, as notebooks will be only added to the list once assumed done?

Yes, I think it would be a mistake to have `make clean` automatically remove unused imports. My preference is that we automate things that developers will almost never care about, like import orders or trailing whitespace, but avoid being too heavy handed beyond that.

Anyway, who would you like to review this @rfl-urbaniak ? I always find it confusing when there are two reviewers listed.

rfl-urbaniak:
Sorry for being confusing. I wanted @eb8680 to approve the general strategy, as this is what would  be applied across all notebooks in the repo, and @SamWitty to approve the resulting changes to the notebook,

SamWitty:
> Sorry for being confusing. I wanted @eb8680 to approve the general strategy, as this is what would be applied across all notebooks in the repo, and @SamWitty to approve the resulting changes to the notebook,

Ok, thanks for clarifying. The changes to the notebook have my stamp of approval. @eb8680 , ball's in your court to review the broader PR.

eb8680 (approved):
Seems OK to me, thanks for taking this on!

PR #533, opened 2024-04-02
Revise tutorial with broadcastable models and counterfactual indexing

SamWitty:
This PR revises our main tutorial to better demonstrate ChiRho's compatibility with tensor broadcasting and indexing counterfactual worlds using `gather`, `IndexSet`, and `indices_of` from `chirho.indexed.ops`.

In addition, this PR removes some lingering TODOs about making sure probabilistic inference is working as expected. I confirmed that SVI is working by adding a new visualization comparing ground_truth ATE to the posterior estimate. There are also some minor changes to phrasing throughout, see changes for detail.

PR #539, opened 2024-04-26
Fix tutorial to pass notebook CI tests

SamWitty:
#528 changed the default name for fresh `split` operations from `intervened` to `__fresh_split__`. It appears that the CI for this PR was run before #533 was merged in, causing the CI to look like it was passing when it actually wasn't. See https://github.com/BasisResearch/chirho/actions/runs/8834817360/job/24259115895 for the error this causes.

This small PR changes any use of `intervened` with `__fresh_split__` in the tutorial notebook, addressing the notebook CI failure.

Edit: This PR now changes `TwinWorldCounterfactual` to `MultiWorldCounterfactual` instead.


eb8680:
Instead of switching to the new correct default name under `TwinWorldCounterfactual`, can you change the tutorial to use `MultiWorldCounterfactual` so that the default name never appears in the first place? I think that would be less confusing.

eb8680 (approved):
Thanks for fixing this!

PR #546, opened 2024-06-20
Fixed bug in "Adding Uncertainty over Model Parameters" section of the tutorial

michaelleejl:
In **Adding Uncertainty over Model Parameters**, (https://basisresearch.github.io/chirho/tutorial_i.html#Adding-Uncertainty-over-Model-Parameters), the following code snippet

```
def bayesian_population_causal_model(n_individuals):
    stress_prob, smokes_cond_prob, cancer_cond_prob = parameter_prior()
    return population_intervened_causal_model(
        n_individuals, stress_prob, smokes_cond_prob, cancer_cond_prob
    )
```

Calls `population_intervened_causal_model`. It should instead call `population_causal_model`.


eb8680 (approved):
Thanks for fixing this!

PR #547, opened 2024-06-27
Update explainable_categorical.ipynb

djinnome:
braces were missing around `m'` such that `P(f_m')` rendered incorrectly as $P(f_m')$ which is equivalent to  $P(F=0 | do(M=1))$ instead of `P(f_{m'})` rendering as $P(f_{m'})$ which is equivalent to $P(F=1|do(M=0))$ as intended


eb8680 (approved):
Thanks for fixing this!

PR #549, opened 2024-07-08
Move cut posterior handlers to robust module

### LINKED ISSUE #495: Cut-posterior handlers should live in robust module
`SingleStageCut`, `CutModule` and `CutComplementModule` currently live in `chirho.observational.handlers.cut`. After #398 lands we should move them to `chirho.robust.handlers.cut` and update the two examples that use them accordingly.

eb8680:
Resolves #495 

This small refactoring PR moves the contents of `chirho.observational.handlers.cut` to `chirho.robust.handlers.cut`, and the associated unit tests from `test/observational/` to `test/robust/`.


SamWitty (approved):
Looks good

PR #550, opened 2024-07-08
Move Predictive handlers to observational module

eb8680:
Blocked by #549

This pure refactoring PR moves some of the utilities for making and batching predictive distributions out of `chirho.robust.handlers.predictive` and `chirho.robust.internals.nmc` and into a new `chirho.observational.handlers.predictive`. There are no changes to the code other than moving definitions to different files (with one small exception, noted below).


chirho/robust/internals/nmc.py
```diff
@@ -205,7 +162,11 @@ def forward(
 
         # move data plate dimension to the left
         for name in reversed(plate_name_to_dim.keys()):
-            log_weights = bind_leftmost_dim(log_weights, name)
+            log_weights = torch.transpose(
```
eb8680, line 165: This is the only non-renaming change in this PR. It is equivalent to the previous version with the definition of `bind_leftmost_dim` inlined for this special case.

SamWitty: I have a slight preference for keeping it as `bind_leftmost_dim`, as that seems a bit more transparent about what the code is doing.

If we do decide to remove it here is `bind_leftmost_dim` used anywhere else anymore?

SamWitty (changes requested):
As noted I have a slight preference to revert the `bind_leftmost_dim` change, but I'll happily defer to your preferences here.

I marked this as "request changes" just because the linter seems to be failing because of unused imports. Once that's resolved I'll approve.

chirho/robust/internals/nmc.py
eb8680: I inlined `bind_leftmost_dim` here because it is now in `observational.internals` and shouldn't be used outside `observational`. I could have moved `BatchedNMCLogMarginalLikelihood`, but I would like to keep that as an internal implementation detail of `chirho.robust` for now.

`bind_leftmost_dim` is still used inside `chirho.observational.handlers.predictive.PredictiveFunctional`, which was moved in this PR.

SamWitty (approved):

PR #551, opened 2024-07-10
Replace bespoke flattening logic with PyTree-based flattening in `chirho.robust` CG solver

eb8680:
This refactoring PR replaces the bespoke `functools.singledispatch`-based flattening/unflattening logic in `chirho.robust.internals.utils.make_flatten_unflatten` with equivalent logic using PyTorch's PyTree `flatten`/`unflatten` implementations, which are already used elsewhere in `chirho.robust`.

This should not change the behavior of `make_flatten_unflatten` and should be exercised by existing unit tests.

### DIFF
--- a/chirho/robust/internals/utils.py
+++ b/chirho/robust/internals/utils.py
@@ -28,9 +28,8 @@ U = TypeVar("U")
 ParamDict = Mapping[str, torch.Tensor]
 
 
-@functools.singledispatch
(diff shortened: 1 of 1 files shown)


azane (approved):
Looks good — that said, we should probably add tests that exercise the more general pytree structures at some point. I think the current ones just do single-level dicts?

PR #552, opened 2024-07-19
TEMP PR for Change Tracking (delete after July 2024)

azane:
Just tracking the changes between these two branches so I can port the changes one-by-one, and make comments accordingly, to the new DiffEqPy backend implementation.


chirho/dynamical/handlers/interruption.py
```diff
@@ -34,7 +34,9 @@ def __init__(self, time: R):
         super().__init__()
 
     def event_fn(self, time: R, state: State[T]) -> R:
-        return cond(0.0, self.time - time, case=time < self.time)
+        # FIXME jd0120dh cond seems backwards to me? But this returns the second val
+        #  where case is true and first if false...
+        return cond(self.time - time, 0.0, case=time >= self.time)
```
azane, line 39: Not sure that this was for...

PR #555, opened 2024-08-02
Instrumental example notebook.

dimkab:
The notebook probably needs a bit more polishing, but I would love comments at this stage.


eb8680 (commented):
Looks great! My main suggestion is to include lines for both the true ATE and the biased naive estimate in the two ATE plots and to make sure both plots have the same scale on the X axis. That way it's clearer that the mean of your estimate is much closer to the true value than the naive estimate.

Other comments:
- Consider adding plots showing posterior predictive fits, e.g. `pairplot`s comparing the true and posterior predictive marginal distributions for each observed variable (`X`, `Y` and `Z`).
- Consider adding section headers corresponding more closely to other notebooks, such as those in the mediation analysis and backdoor adjustment examples. The more structurally similar this example is to the others and to the systematic workflow in the final section of the tutorial, the easier it will be to understand for someone who's already read those.
- Make sure to build (`cd docs && make html`) and inspect (`python -m http.server -d docs/build/html/ <port>` and open `https://localhost:<port>` in a browser) the docs locally to verify that things render correctly. It looks to me like the line wrapping of equation 1 might be messed up.

eb8680:
Oh, and also please add this to the `Examples` section in `docs/source/index.rst` and the "Learn more" section of `README.rst`, otherwise it won't actually be included in the docs.

dimkab:
Thanks! I've  followed all your suggestions, please see revised version.

eb8680 (approved):
Looks good to me!

PR #557, opened 2024-08-06
Explainability edge cases

PoorvaGarg:
This PR targets how Chirho's explainable module handles cases where the `antecedents` are independent of `consequents` or have no causal role, while ensuring the cases of dependence are still handled properly. The attempts to fix a bug led to code changes mainly in `consequent_eq_neq`. Extensive tests closely related to the bug, defined over two and three variable models have been added in `test_handlers_explainable`  to ensure expected behavior both for independence and causal connection.


rfl-urbaniak:
@eb8680 We pair-coded most of the fixes, so it's better someone else takes a fresh look.

tests/explainable/test_handlers_components.py
```diff
@@ -376,49 +376,64 @@ def model_ce():
 def test_consequent_eq_neq(plate_size, event_shape):
     factors = {
         "consequent": consequent_eq_neq(
-            support=constraints.independent(constraints.real, len(event_shape)),
-            proposed_consequent=torch.Tensor([0.1]),  # added this
+            support=constraints.independent(constraints.real, 0),
```
eb8680, line 379: This `constraint` should still have `event_dim` of `len(event_shape)`.

tests/explainable/test_handlers_components.py
```diff
@@ -376,49 +376,64 @@ def model_ce():
...
+            proposed_consequent=torch.Tensor([0.01]),
```
eb8680, line 380: I believe `proposed_consequent` should have shape `event_shape`.

tests/explainable/test_handlers_components.py
```diff
@@ -376,49 +376,64 @@ def model_ce():
...
     def model_ce():
-        w = pyro.sample(
-            "w", dist.Normal(0, 0.1).expand(event_shape).to_event(len(event_shape))
-        )
-        consequent = pyro.deterministic(
-            "consequent", w * 0.1, event_dim=len(event_shape)
-        )
-
-        return consequent
+        w = pyro.sample("w", dist.Normal(w_initial, 0.001))
```
eb8680, line 392: This distribution should still have shape `event_shape`.

tests/explainable/test_handlers_components.py
```diff
@@ -376,49 +376,64 @@ def model_ce():
...
+        consequent = pyro.deterministic("consequent", w * torch.tensor(0.1))
```
eb8680, line 393: This `deterministic` site should still have `event_dim` of `len(event_shape)`.

tests/explainable/test_handlers_components.py
```diff
@@ -376,49 +376,64 @@ def model_ce():
...
+        eq_neq_log_probs_fact = gather(
+            nd["__factor_consequent"]["fn"].log_factor,
+            IndexSet(**{"w": {0}}, event_dim=0),
+        )
+        eq_neq_log_probs_nec = gather(
+            nd["__factor_consequent"]["fn"].log_factor,
+            IndexSet(**{"w": {1}}, event_dim=0),
+        )
+        consequent_suff = gather(
+            nd["consequent"]["value"], IndexSet(**{"w": {2}}, event_dim=0)
```
eb8680, line 422: I believe this `gather` call should have `event_dim` of `len(event_shape)`.

tests/explainable/test_handlers_components.py
```diff
@@ -376,49 +376,64 @@ def model_ce():
 def test_consequent_eq_neq(plate_size, event_shape):
```
eb8680, line 376: All the logic about `event_shape` and `event_dim` in this test needs to be restored, see other comments.

chirho/explainable/handlers/components.py
```diff
@@ -292,16 +278,13 @@ def _consequent_eq_neq(consequent: T) -> torch.Tensor:
...
         nec_suff_log_probs_partitioned = {
+            **{null_index: FACTUAL_NEC_SUFF},
             **{
-                factual_indices: FACTUAL_NEC_SUFF,
-            },
-            **{
-                IndexSet(**{antecedent: {ind}}): log_prob
-                for antecedent in (
-                    set(antecedents)
-                    & set(indices_of(consequent, event_dim=support.event_dim))
```
eb8680, line 303: Was this line restricting the index variables to just those already present in `consequent` the root cause of a problem? If so, which of the new or updated tests in this PR fail with this condition in place and pass with the changes?

tests/explainable/test_handlers_explanation.py
```diff
@@ -314,3 +316,236 @@ def test_SplitSubsets_two_layers():
...
+            antecedent_bias=-0.5,
+            consequent_scale=0,
+        ):
+            with pyro.plate("sample", size=3):
+                with pyro.poutine.trace() as trace_reverse:
+                    model_connected()
+
+    trace_connected.trace.compute_log_prob
+    trace_independent.trace.compute_log_prob
+    trace_reverse.trace.compute_log_prob
```
eb8680, line 382: `compute_log_prob` is a method, not a property, so you need to call it (`compute_log_prob()`) in these lines for them to do anything.

tests/explainable/test_handlers_explanation.py
```diff
@@ -314,3 +316,236 @@ def test_SplitSubsets_two_layers():
         ).item()
 
     assert obs_bill_hits == 0.0 and int_bill_hits == 0.0 and int_bottle_shatters == 0.0
+
+
+def test_edge_eq_neq():
```
eb8680, line 321: Looks like the two models in this test don't actually interact in any of the assertions? If that's the case, consider splitting this into two or more separate tests, one for each model and conditioning set, as you've already done for `test_eq_neq_three_variables` below.

tests/explainable/test_handlers_explanation.py
```diff
@@ -314,3 +316,236 @@ def test_SplitSubsets_two_layers():
...
+                with pyro.poutine.trace() as trace:
+                    model()
+
+    trace.trace.compute_log_prob
+    nodes = trace.trace.nodes
+
+    values = nodes[cons]["value"]
+    log_probs = nodes[f"__cause____consequent_{cons}"]["fn"].log_factor
+
+    assert log_probs.shape == torch.Size([3, 3, 1, 1, 1, 1])
```
eb8680, line 534: Consider asserting something about `indices_of(log_probs)` instead of `log_probs.shape` here. The shape is an implementation detail of `IndexPlatesMessenger` / `MultiWorldCounterfactual` and may change.

tests/explainable/test_handlers_explanation.py
```diff
@@ -314,3 +316,236 @@ def test_SplitSubsets_two_layers():
...
+    [
+        model_three_converge,
+        model_three_diverge,
+        model_three_chain,
+        model_three_complete,
+        model_three_isolate,
+        model_three_independent,
+    ],
+)
+def test_eq_neq_three_variables(model, ante_cons):
```
eb8680, line 509: To make this test even more robust, you might consider additionally parametrizing it by an `event_shape` for the model variables.

tests/explainable/test_handlers_explanation.py
```diff
@@ -314,3 +316,236 @@ def test_SplitSubsets_two_layers():
...
+    trace_independent.trace.compute_log_prob
+    trace_reverse.trace.compute_log_prob
+
+    Y_values_ind = trace_independent.trace.nodes["Y"]["value"]
+
+    log_probs_ind = trace_independent.trace.nodes["__cause____consequent_Y"][
+        "fn"
+    ].log_factor
+    if torch.any(Y_values_ind == 1.0):
+        assert log_probs_ind[1, 0, 0, 0, :].sum().exp() == 0.0
```
eb8680, line 390: All of these indexing operations should be replaced with equivalent calls to `gather` instead of hardcoding batch shapes, which are implementation details of `MultiWorldCounterfactual` and may change.

tests/explainable/test_handlers_explanation.py
```diff
@@ -314,3 +316,236 @@ def test_SplitSubsets_two_layers():
...
+            witnesses={},
+            alternatives={ante1: torch.tensor(0.0), ante2: torch.tensor(0.0)},
+            antecedent_bias=-0.5,
+            consequent_scale=0,
+        ):
+            with pyro.plate("sample", size=1):
+                with pyro.poutine.trace() as trace:
+                    model()
+
+    trace.trace.compute_log_prob
```
eb8680, line 528: `Trace.compute_log_prob` is a method rather than a `property` and must be called (`compute_log_prob()`) for this line to do anything.

eb8680 (commented):
Thanks for adding these new tests!

tests/explainable/test_handlers_components.py
PoorvaGarg: The logic for `event_shape` has been restored.

eb8680 (approved):
Looks good!

PR #578, opened 2025-02-21
Introduce sampling dependency for auxiliary nodes in `SearchForExplanation`

rfl-urbaniak:
Search cases where antecedent and witness candidates overlap occur fairly naturally when we suspect a node to be a cause and we also want to allow it to be a witness when testing other nodes as antecedents (same run forward, different samples). Ideally, we should be able to do this in a single search. 

Currently, however, witness and antecedent auxiliary nodes are sampled independently. This, among other things, means that if an antecedent candidate is not preempted (i.e. we intend to execute an antecedent intervention) the witness auxiliary node can still be one (stochastically), effectively preempting this antecedent intervention. 

This is not the intended behavior. We know that if a node is considered as an antecedent, it should not be at the same time treated as a potential witness, and so the intended behavior should be that witness sampling is reserved for samples in which the antecedent intervention is not performed, effectively setting witness preemption `case` to 0 if the antecedent preemption case is `0`. Not doing so not only leads to inefficient sampling, but also might skew the probability estimates if they involve conditioning on the antecedent intervention being performed but the user forgets to condition at the same time on that node not being a witness.

This PR allows for improved behavior by pulling case sampling outside of the handlers and passing the sampled and dependency-corrected cases forward to the handlers. 

The API is backward compatible - if `num_samples` and `sampling_dimension` are not passed, we default to the old behavior, which is sufficient for the more limited search cases that we have been using so far.


rfl-urbaniak:
There might be still some unrelated linting issues with the dynamical module, which I think will be resolved independently.

SamWitty:
@eb8680 , I believe this branch will pass once rebased to `master`, as the linting error was previously resolved by https://github.com/BasisResearch/chirho/pull/581

rfl-urbaniak:
Updated from master, resolved linting errors.

chirho/explainable/handlers/explanation.py
```diff
@@ -85,13 +91,20 @@ def SearchForExplanation(
...
     :param preemptions: An optional mapping of names to witness preemption values.
-    :param antecedent_bias: The scalar bias towards not intervening. Must be between -0.5 and 0.5, defaults to 0.0.
-    :param consequent_scale: The scale of the consequent factor functions, defaults to 1e-2.
-    :param witness_bias: The scalar bias towards not preempting. Must be between -0.5 and 0.5, defaults to 0.0.
+    :param antecedent_bias: A scalar bias towards not intervening. Must be between -0.5 and 0.5. Defaults to 0.0.
+    :param consequent_scale: The scale of the consequent factor functions. Defaults to 1e-2.
+    :param witness_bias: A scalar bias towards not preempting. Must be between -0.5 and 0.5. Defaults to 0.0.
     :param prefix: A prefix used for naming additional consequent nodes. Defaults to ``__consequent_``.
+    :param num_samples: The number of samples to be drawn for each antecedent and witness.
+                        Needed if witness and antecedent samples are to be coordinated.
```
eb8680, line 99: I can't tell from the code or this rather mysterious fragment of documentation what either of the parameters `num_samples` or `sampling_dim` are supposed to achieve or what effect different values are supposed to have on the behavior of `SearchForExplanation`, especially as the other parameters are varied.

Maybe that's on me, but since I know something about the inner workings of `SearchForExplanation`, I think it's reasonable to assume other people who don't will have an even harder time understanding. I'm also not sure this is a solvable problem with documentation alone - the interface for `SearchForExplanation` is already so complex that there's really no room to add more choice to it.

chirho/explainable/handlers/explanation.py
```diff
@@ -85,13 +91,20 @@ def SearchForExplanation(
...
+    :param sampling_dim: The dimension along which the antecedent and witness nodes will be sampled,
+                        to be kept consistent with case sampling.
+
+    :note: If ``num_samples`` is not provided, the antecedent and witness nodes will be sampled independently.
```
eb8680, line 103: Isn't it even more important to say what happens when a value for this new parameter is provided?

tests/explainable/test_handlers_explanation.py
```diff
@@ -191,6 +210,48 @@ def test_SearchForExplanation():
...
+            consequents=consequents,
+            witnesses=witnesses,
+            alternatives=alternatives,
+            antecedent_bias=0.1,
+            consequent_scale=1e-8,
+            num_samples=100,
+            sampling_dim=-1,
+        ):
+            with observations_conditioning:
+                with pyro.plate("sample", 100):
```
eb8680, line 238: What is this plate for? Is it supposed to work with the `num_samples` argument, or is it meant to be unrelated?

chirho/explainable/handlers/preemptions.py
```diff
@@ -53,11 +58,13 @@ def __init__(
         *,
         prefix: str = "__witness_split_",
         bias: float = 0.0,
+        cases: Optional[Mapping[str, torch.Tensor]] = None,
```
eb8680, line 61: As with `SearchForExplanation`, it's hard for me (and presumably anyone less familiar with the code) to understand this complex new piece of the `Preemptions` interface, especially given the brief and cryptic documentation provided. Is there more than one use case for it, or even more than one correct set of values in the context of a given `SearchForExplanation` handler? If not, shouldn't it be our job to deal with that behind the scenes and avoid exposing it to users entirely?

chirho/explainable/handlers/explanation.py
```diff
@@ -131,12 +148,43 @@ def SearchForExplanation(
...
+        case_shape[sampling_dim] = num_samples
+
+        antecedent_probs = torch.tensor(
+            [0.5 - antecedent_bias] + ([0.5 + antecedent_bias])
+        )
+
+        antecedent_case_dist = dist.Categorical(probs=antecedent_probs)
+
+        antecedent_cases = {
+            key: antecedent_case_dist.sample(case_shape) for key in antecedents.keys()
```
eb8680, line 165: A Pyro program should never be calling `Distribution.sample` directly and should never introduce batch dimensions that do not correspond to a `plate` context, otherwise the program's meaning and the correctness of any inference algorithm applied to it are undefined.

chirho/explainable/handlers/explanation.py
```diff
@@ -131,12 +148,43 @@ def SearchForExplanation(
...
+        }
+
+        witness_probs = torch.tensor([0.5 - witness_bias] + ([0.5 + witness_bias]))
+
+        witness_case_dist = dist.Categorical(probs=witness_probs)
+
+        witness_cases = {
+            key: witness_case_dist.sample(case_shape) for key in witnesses.keys()
```
eb8680, line 173: Ditto: it is never correct to call `Distribution.sample` or introduce unhandled batch dimensions in a Pyro program.

rfl-urbaniak:
closing as superseded by #587

PR #583, opened 2025-03-30
Fix TOCs in two dynamical systems notebooks

### LINKED ISSUE #582: Titles and tables of contents in notebooks render incorrectly in Sphinx
The notebooks added in #565 and #542 render incorrectly in Sphinx, as in this screenshot of the most recent GitHub Pages deployment:

Image

rfl-urbaniak:
Resolves #582. Resulted from the TOCs being generated automatically, which led to some HTML presence in the TOCs, which interacted with Sphinx in an unexpected manner.


rfl-urbaniak:
will be overriden in a different PR.

PR #587, opened 2025-04-10
prioritize active interventions over witness preemptions

rfl-urbaniak:


chirho/explainable/handlers/explanation.py
```diff
@@ -61,7 +61,7 @@ def SearchForExplanation(
     *,
     alternatives: Optional[Mapping[str, Intervention[S]]] = None,
     factors: Optional[Mapping[str, Callable[[T], torch.Tensor]]] = None,
-    preemptions: Optional[Mapping[str, Union[Intervention[S], Intervention[T]]]] = None,
+    # preemptions: Optional[Mapping[str, Union[Intervention[S], Intervention[T]]]] = None,
```
eb8680, line 64: Can you revert the change deleting this argument? I don't think it's correct and in any case it's not related to the other change that swaps the orders of the witness and antecedent handlers.

chirho/explainable/handlers/explanation.py
```diff
@@ -81,15 +81,18 @@ def SearchForExplanation(
     :param supports: A mapping of sites to their support constraints.
     :param antecedents: A mapping of antecedent names to optional observations.
     :param consequents: A mapping of consequent names to optional observations.
-    :param witnesses: A mapping of witness names to optional observations.
+    :param witnesses: A mapping of witness names to optional observations or custom preemption values.
```
eb8680, line 84: The `witnesses` and `preemptions` arguments are different things and cannot be combined in this way, as their types would seem to imply and as is true for the `antecedents` and `alternatives` arguments. Like `antecedents`, `witnesses` contains (optional) observed values for witnesses in the factual world. Like `alternatives`, `preemptions` contains interventions applied when a witness is preempted.

Values provided in `witnesses` are simply passed through the mapping yielded by `SearchForExplanation` to facilitate separate enforcement of the factivity condition.

eb8680 (changes requested):
Please revert the removal of `preemptions`

chirho/explainable/handlers/explanation.py
rfl-urbaniak: Reinstated and modified the test accordingly. I do still think there is some redundancy in the witness values if passed not being used anywhere. But you're right; this is not directly connected to this change.

rfl-urbaniak: One point worth noticing perhaps is that with this change of the order of handlers, witness preemptions still change the sample, but only in the factual world, whereas the counterfactual worlds follow the antecedent interventions. It is not clearly undesirable and makes a difference only in rather strange circumstances when one passes non-factual preemptions. As in, say, row 4 here:

image

rfl-urbaniak: where the factual value is 20, the alternative is drawn from the proposal, age_i is age antecedent intervention preemption, and age_w is age witness preemption

eb8680 (approved):
Looks good now, can this be merged or is there something you still want to add?

rfl-urbaniak:
Yes, it's ready to get merged. I will likely slightly modify the construction of the antecedent interventions, but that's for another PR. Thank you!

PR #589, opened 2025-04-15
basic setup for the COMPAS notebook

rfl-urbaniak:
I propose we proceed in smaller stages when building the COMPAS case study notebook. This, hopefully, will speed up the feedback process and facilitate alignment as we move forward. Such steps will be collected in `staging-compas`. For now, the first part introduces the problem, the surrogate, and `SearchForExplanation` sampling to the reader.


rfl-urbaniak:
closing as no longer relevant

PR #594, opened 2025-09-10
Add a batched intervention handler

jfeser:
In search for explanation, we need to evaluate large batches of interventions. These interventions are not homogenous---they may intervene on different variables. This handler evaluates batches of interventions efficiently. It uses `chirho.indexed` to create a new dimension to store the results.


rfl-urbaniak:
closing as superseded

PR #598, opened 2025-09-15
Port to pyproject and uv

jfeser:


jfeser:
I think the notebook failure is spurious.

eb8680:
New failures after #595 landed need to be addressed, then this can land.

eb8680:
> I think the notebook failure is spurious.

Seems otherwise, there's a different error now.

PR #601, opened 2025-10-08
Reparametrized normal with base noise exposed

rfl-urbaniak:


chirho/observational/reparams.py
```diff
@@ -0,0 +1,47 @@
...
+from pyro.infer.reparam.reparam import Reparam
+
+from chirho.indexed.ops import IndexSet, indices_of
+
+
+class NormalReparam(Reparam):
+    def __init__(
+        self,
+    ):
+        super().__init__()
```
jfeser, line 13: Remove me.

chirho/observational/reparams.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def apply(self, msg):
+        name = msg["name"]
+        fn = msg["fn"]
+        value = msg["value"]
+        is_observed = msg["is_observed"]
+
+        if isinstance(fn, dist.Independent):
+            base = fn.base_dist
```
jfeser, line 22: Are we assuming that all `Independent` wrapped dists are `Normal`?

chirho/observational/reparams.py
```diff
@@ -0,0 +1,47 @@
...
+from chirho.indexed.ops import IndexSet, indices_of
+
+
+class NormalReparam(Reparam):
+    def __init__(
+        self,
+    ):
+        super().__init__()
+
+    def apply(self, msg):
```
jfeser, line 15: `def apply(self, msg: pyro.infer.reparam.reparam.ReparamMessage) -> pyro.infer.reparam.reparam.ReparamResult`

chirho/observational/reparams.py
```diff
@@ -0,0 +1,47 @@
...
+
+        if is_observed:
+            value_indices = indices_of(value)
+            if value_indices == IndexSet():
+                return {"fn": fn, "value": value, "is_observed": is_observed}
+
+            if value_indices != IndexSet():
+                raise NotImplementedError("Partially observed Normal reparameterization is not implemented.")
+
+        if not is_observed:
```
jfeser, line 40: `else:`

jfeser (changes requested):

chirho/observational/reparams.py
rfl-urbaniak: Those to which you specifically point as the ones you want to reparametrize as reparametrized normal, I think so. Any reason to think otherwise?

jfeser: I think it's worth a check since reparam can be used automatically as well.

chirho/observational/reparams.py
```diff
@@ -0,0 +1,47 @@
...
+        elif isinstance(fn, dist.Normal):
+            base = fn
+        else:
+            raise ValueError(f"NormalReparam only supports Normal or Independent(Normal), got {type(fn)}")
+
+        loc = base.loc
+        scale = base.scale
+        event_dim = fn.event_dim
+
+        if is_observed:
```
dimkab, line 32: I suggest to raise an exception here. Reparametrization is not supposed to be used on observed sites, and in fact none of the tests do so. See e.g. 
https://docs.pyro.ai/en/1.5.0/_modules/pyro/infer/reparam/loc_scale.html

tests/observational/test_reparametrized.py
```diff
@@ -0,0 +1,267 @@
...
+            with pyro.poutine.trace() as tr_conditioned:
+                small_model()
+
+    tr_conditioned.trace.compute_log_prob()
+    y2 = tr_conditioned.trace.nodes["y"]
+    assert torch.allclose(y2["value"], y_value)
+    assert torch.allclose(y2["log_prob"], dist.Normal(loc, scale).log_prob(y2["value"]))
+
+    # reparam will encouter Delta not a Normal or Independent(Normal)
+    with pytest.raises(ValueError, match="NormalReparam only supports Normal or Independent\\(Normal\\)"):
```
dimkab, line 84: again, reparam is not supposed to be used on observed sites, and therefore this part of the test is not needed

dimkab (changes requested):
reparam is not supposed to be used on observed sites. better to raise an exception.

rfl-urbaniak:
closing as superseded

PR #604, opened 2026-04-13
Give excision errors an exception subtype

jfeser:
We catch these elsewhere and matching on the message is fragile.


rfl-urbaniak:
@jfeser hey, what's the status of this work?

jfeser:
It's not important. We can close.

PR #606, opened 2026-07-06
batched interventions with split

rfl-urbaniak:
## 1. Motivation

Many causal workflows, explanation search in particular, require evaluating
large batches of heterogeneous interventions: different scenarios may intervene
on different variables with different values.

ChiRho has two counterfactual handlers:

- **`TwinWorldCounterfactual` (TWC)**: factual + one alternative. One shared
  axis of size 2. Cheap, but limited to a single counterfactual.

- **`MultiWorldCounterfactual` (MWC)**: factual + every combination. Each
  intervened site gets its own axis, so the result covers the full Cartesian
  product. Memory grows exponentially in the number of sites.

The missing point: N heterogeneous scenarios on a single shared axis. Linear
memory, one forward pass, no combinations nobody asked for.

## 2. Semantics

For N scenarios, the world axis has size N+1:

- **Index 0** — factual world, no interventions applied.
- **Index i+1** — scenario i: at site s, the value is `act[s,i]` if scenario i
  intervenes on s, otherwise the propagated value from upstream interventions
  in that scenario (factual if none).

Each batched world equals the corresponding MWC slice when interventions target a single site. 
With multiple intervened sites the noise draws differ across cells. The worlds are distributionally equivalent
but not sample-path identical.

## 3. Requirements

1. One shared axis of size N+1 (linear in N, not exponential in sites).
2. One vectorized forward pass.
3. Non-intervened sites reflect upstream interventions within the same scenario,
   not the global factual.
4. Factual world always at index 0.
5. Non-trivial event dimensions (`to_event(k)`) work correctly.
6. Standard `chirho.indexed` interface: `gather`, `IndexSet`, `indices_of`.
7. Pre-sampled `(act, mask)` tensors can be passed directly to `do` — no
   Python loop needed to construct scenarios.

## Composition with MWC

Nesting `BatchedWorldCounterfactual` inside `MultiWorldCounterfactual` (or any
other `IndexPlatesMessenger` subclass) is not currently supported. Both handlers
run `_pyro_add_indices` for the same plate name and try to enter a Pyro plate
with the same `__index_plate___<site>` name, which raises a `ValueError`.
This is the same pre-existing limitation as nesting MWC inside
`TwinWorldCounterfactual`, not specific to BWC.

A plain `do()` without an `IndexPlatesMessenger` companion (e.g. fixing one site
to a constant while BWC runs) does work: BWC's `_pyro_intervene` skips
non-`BatchedAction` interventions, so the plain action is applied as a scalar
that broadcasts across the `batched_interventions` axis.

### Shared Noise

Unintervened sites keep their single scalar draw, which broadcasts across the
world axis. 

## 5. API

**`BatchedAction(act, mask=None)`**

- `act`: `(N, *event_shape)` — intervention value for each scenario.
- `mask`: `(N,)` bool — which scenarios intervene on this site. Defaults to
  all-True.

Slots where `mask=False` are never read by `torch.where`; their values are irrelevant placeholders.
`batch_scenarios` fills them with zeros; when constructing `BatchedAction` directly the caller
may put anything there.

**`BatchedWorldCounterfactual(first_available_dim=None)`**

Subclasses `IndexPlatesMessenger` and `BaseCounterfactualMessenger`, intercepts `split` for `BatchedAction` values.

**`batch_scenarios(*scenarios)`**

Helper function. Converts a sequence of `{site: tensor}` dicts
into `{site: BatchedAction}` by inferring per-site masks from key presence.
Users with pre-sampled tensors can build the dict directly without it.

Results:

```python
gather(value, IndexSet(batched_interventions={k}), event_dim=...)
# k=0 is factual, k=1..N are the scenarios
```

---

### Use case 1: sweep one site over N values

```python
z_vals = torch.linspace(-3.0, 3.0, 5)  # shape (5,)

with BatchedWorldCounterfactual():
    with do(actions={"z": BatchedAction(act=z_vals)}):
        z, x, y = model()

# world 0: factual z draw
# worlds 1-5: z fixed at each value; all worlds share the same noise elsewhere
```

No mask needed — omitting it means every scenario intervenes.

---

### Use case 2: heterogeneous masks across sites

Three scenarios, each intervening on a different subset of sites:

```python
actions = {
    "z": BatchedAction(
        act=torch.tensor([1.0, 2.0, 0.0]),   # slot 2 is a placeholder
        mask=torch.tensor([True, True, False]),
    ),
    "x": BatchedAction(
        act=torch.tensor([0.0, 3.0, 3.0]),   # slot 0 is a placeholder
        mask=torch.tensor([False, True, True]),
    ),
}

with BatchedWorldCounterfactual():
    with do(actions=actions):
        z, x, y = model()

# world 0: factual
# world 1: z=1 only; x propagates from modified z
# world 2: z=2 and x=3 simultaneously
# world 3: x=3 only; z is the factual draw
```

With MWC the same setup produces four worlds across two axes (Cartesian product).
Here all three scenarios share one axis.

---

### Use case 3: scenario dicts via `batch_scenarios`

```python
with BatchedWorldCounterfactual():
    with do(actions=batch_scenarios(
        {"z": torch.tensor(1.0)},
        {"z": torch.tensor(2.0), "x": torch.tensor(3.0)},
        {"x": torch.tensor(3.0)},
    )):
        z, x, y = model()
```

Same result as use case 2. `batch_scenarios` infers masks from which keys appear
in each dict.

---

#### Background

PCI (probabilistic causal identification) needs to run N pairs of model
evaluations — one sufficiency world (factual values at active sites) and one
necessity world (alternative values at same sites) — per trial. Currently this
is a Python loop: the model is called twice per iteration and noise is manually
threaded from the first run into the second.

The loop can be replaced with a single batched call. Masks (which sites are
active per trial) and values (what to set them to) are sampled up front as
tensors and assembled into `BatchedAction` objects. The 2N scenarios run in one
pass; noise is shared automatically across all worlds, so each (sufficiency,
necessity) pair sees the same exogenous draws without any manual extraction.

This is only possible with the `(act, mask)` tensor interface. Scenario dicts
require every site assignment to be fixed at construction time, which rules out
the vectorised pre-sampling pattern.

#### Example

```python
N = 4
sites = ["z", "x"]

# Pre-sample which sites are active per scenario (from sample_k_indices)
masks = {
    "z": torch.tensor([True,  False, True,  False]),
    "x": torch.tensor([False, True,  True,  False]),
}

# Pre-sample intervention values; slots where mask=False are placeholders
suff_vals = {"z": torch.full((N,), 5.0),  "x": torch.full((N,),  5.0)}
nec_vals  = {"z": torch.full((N,), -5.0), "x": torch.full((N,), -5.0)}

# 2N scenarios: sufficiency first, necessity second
actions = {
    site: BatchedAction(
        act=torch.cat([suff_vals[site], nec_vals[site]]),
        mask=torch.cat([masks[site],    masks[site]]),
    )
    for site in sites
}

with BatchedWorldCounterfactual():
    with do(actions=actions):
        z, x, y = model()

# world 0:      factual
# worlds 1..N:  sufficiency (factual values at active sites)
# worlds N+1..: necessity   (alternative values at same sites)

suff_y = [gather(y, IndexSet(batched_interventions={i})).reshape(()) for i in range(1, N+1)]
nec_y  = [gather(y, IndexSet(batched_interventions={i})).reshape(()) for i in range(N+1, 2*N+1)]
```

## 6. Benchmarks

Run locally, single thread, skipped in CI.

**Wall-clock time**, N=300 single-site interventions, three-site model, event size 300:

| Approach | Time | vs batched |
|---|---|---|
| this handler | 1.2 ms | 1x |
| MWC | 28.5 ms | 25x |
| loop + `do` | 37.8 ms | 33x |
| loop + TWC | 195.5 ms | 170x |

**Output tensor size** as the number of intervened sites K grows:

| K | MWC bytes | batched bytes | ratio |
|---|---|---|---|
| 2 | 16 | 12 | ~1x |
| 4 | 64 | 20 | ~3x |
| 8 | 1024 | 36 | ~28x |
| 12 | 16384 | 52 | ~315x |

## 7. Prior work

[#594](https://github.com/BasisResearch/chirho/pull/594) (Jack Feser): the
original prototype. Batch interventions on a shared leading dimension, select
with `torch.where`. Two bugs: the stacking loop read the loop variable instead
of the accumulator (only the last scenario's values were used), and the batch
dimension was a raw tensor axis invisible to `chirho.indexed`. No working handler
implementation.

[#605](https://github.com/BasisResearch/chirho/pull/605): fixed both bugs,
registered the batch dimension as a named index plate, added a factual world at
index 0, and a `shared_noise` flag for MWC-equivalent counterfactuals.

This PR follows Jack's review of #605. He pointed out that intercepting
`intervene` rather than `split` puts the handler outside the normal counterfactual
machinery, that passing interventions through the constructor rather than `do`
breaks composability, and that `shared_noise` should not be a user-facing flag.
The main change here is moving to `_pyro_split` and subclassing
`BaseCounterfactualMessenger`, which resolves all three. `_BatchedAction` becomes
the public `BatchedAction` so users with pre-sampled tensors can construct
scenarios directly without going through scenario dicts.


jfeser (approved):
This interface looks reasonable to me. I've pushed a commit that simplifies the type of `BatchedAction` and removes a type ignore.

eb8680 (approved):

PR #607, opened 2026-07-22
Differentiable masks for `BatchedAction`

rfl-urbaniak:
`BatchedAction.mask` now accepts a floating point tensor in addition to `bool`. A bool mask
gates through `torch.where` exactly as before. A float mask `w` blends against the site's
realized value:

```python
if mask.dtype is torch.bool:
    msg["value"] = torch.where(mask, act_value, obs)   # unchanged
else:
    w = mask.to(obs.dtype)
    msg["value"] = w * act_value + (1 - w) * obs
```

`torch.where` blocks learning *which* sites to intervene on by gradient descent, 
which is the one thing the batched machinery is otherwise well suited to.

Autograd rejects `requires_grad` on bool, so we use a floating point dtype.
A float mask holding exactly 0.0/1.0 is still a hard gate, 
and that is the intended use, typically via a straight-through estimator.

### DIFF
--- a/chirho/counterfactual/handlers/counterfactual.py
+++ b/chirho/counterfactual/handlers/counterfactual.py
@@ -206,78 +206,74 @@ _DEFAULT_BATCH_NAME = "batched_interventions"
 
 @dataclasses.dataclass
 class BatchedAction:
-    """Per-site batched intervention for use with :class:`BatchedWorldCounterfactual`.
+    """Per-site batched intervention for :class:`BatchedWorldCounterfactual`.
 
-    :param act: Shape ``(N, *event_shape)``. Intervention value for each scenario.
-    :param mask: Shape ``(N,)``, bool or floating point. Which scenarios intervene on
-        this site. Defaults to all-True.
-    :param validate_args: If True, check that a floating point ``mask`` lies in ``[0, 1]``.
+    :param act: Intervention values, shape ``(*batch, N, *event_shape)``.
+        ``N = act.shape[-(event_dim + 1)]``.
+    :param mask: Shape ``(*batch, N)``, bool or float. Selects which scenarios
+        intervene on this site. Defaults to all-True.
+    :param event_dim: Event dimensions trailing N in ``act``. Default 0.
 
-    A bool ``mask`` gates. ``torch.where`` never reads slots where it is ``False``,
-    so their ``act`` values are irrelevant placeholders.
+    A bool mask gates: ``False`` slots return the factual value, their ``act``
+    entries ignored.
 
-    A floating point ``mask`` ``w`` blends, ``w * act + (1 - w) * obs``, and passes
-    gradient to both ``act`` and ``w``. Autograd rejects ``requires_grad`` on bool, so
-    a floating point dtype is the only way to learn a mask. Values of exactly 0 or 1
-    match the bool path bit for bit; that is the intended use, typically via a
-    straight-through estimator. Intermediate values give a soft intervention, and
-    identification results that assume atomic interventions no longer hold. Values
-    outside ``[0, 1]`` extrapolate. The blend multiplies both branches instead of
-    picking one, so it requires a floating point site and finite ``act`` and ``obs``.
-    Integer masks raise.
+    A float mask ``w`` blends: ``w * act + (1 - w) * obs``, with gradient flowing
+    to both ``act`` and ``w``. Values 0 and 1 are bit-identical to the bool gate.
+    The intended pattern is a straight-through estimator: a hard 0/1 mask on the
+    forward pass with gradient flowing back through the soft branch to upstream
+    log-probabilities. Intermediate values give a soft blend; identification results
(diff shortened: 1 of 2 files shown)


chirho/counterfactual/handlers/counterfactual.py
```diff
@@ -209,28 +209,51 @@ class BatchedAction:
...
     :param act: Shape ``(N, *event_shape)``. Intervention value for each scenario.
-    :param mask: Shape ``(N,)`` bool. Which scenarios intervene on this site.
-        Defaults to all-True.
-
-    Slots where ``mask`` is ``False`` are never read by ``torch.where``; their
-    values are irrelevant placeholders. All ``BatchedAction`` objects in a single
-    run must share the same leading size N.
+    :param mask: Shape ``(N,)``, bool or floating point. Which scenarios intervene on
+        this site. Defaults to all-True.
+    :param validate_args: If True, check that a floating point ``mask`` lies in ``[0, 1]``.
```
jfeser: I'd just `torch.sigmoid` the mask unconditionally. This check won't work when optimizing anyway.

chirho/counterfactual/handlers/counterfactual.py
```diff
@@ -209,28 +209,51 @@ class BatchedAction:
...
+    def __init__(
+        self,
+        act: torch.Tensor,
+        mask: torch.Tensor | None = None,
+        *,
+        validate_args: bool = False,
+    ):
         self.act = act
         if mask is None:
             self.mask = torch.ones(self.act.shape[0], dtype=torch.bool)
```
jfeser: Use `torch.ones_like` so that the mask will be created on the right device.

chirho/counterfactual/handlers/counterfactual.py
```diff
@@ -209,28 +209,51 @@ class BatchedAction:
...
-        else:
-            if self.act.shape[0] != mask.shape[0]:
-                raise ValueError(
-                    f"act and mask must have the same leading dimension, "
-                    f"got act.shape[0]={self.act.shape[0]} and mask.shape[0]={mask.shape[0]}."
-                )
-            self.mask = mask
+            return
+
+        if self.act.shape[0] != mask.shape[0]:
```
jfeser: Do we want to require masks to be 1-dimensional? What happens if they have trailing dimensions?

jfeser (changes requested):

chirho/counterfactual/handlers/counterfactual.py
rfl-urbaniak: `ones_like(self.act)` produces shape `(N, *event_shape)`, but the mask must be `(N,)` because it gates entire scenarios, not individual elements. But the point about device is sound. Fixed by using `self.act.new_ones(self.act.shape[:1], dtype=torch.bool)`, which creates a `(N,)` bool tensor on the same device as `act`. This broadcasts correctly over any upstream batch dimensions the caller may add to the mask. `test_default_mask_inherits_device` (parametrized over EVENT_SHAPES) verifies device inheritance for non-trivial `act` shapes; `test_batch_mask_value_correctness` and `test_batch_mask_gradient_correctness` verify that `(*batch, N)` masks produce correct per-batch-element values and analytic gradients.

chirho/counterfactual/handlers/counterfactual.py
rfl-urbaniak: Generalized shape considerations inspired by this.
Masks are `(*batch, N)` with N always last; `act` is `(*batch, N, *event_shape)` with
N at `act.shape[-(event_dim + 1)]`. `event_dim` (default 0) is stored on `BatchedAction`,
used in `_prepend_factual_world` (cats on the N axis via `torch.narrow`) and in
`_pyro_split` (`movedim(-(event_dim+1), 0)` before `unbind_leftmost_dim`).
`batch_scenarios` infers `event_dim` from the broadcasted tensors.

`test_mask_batch_shapes_accepted`, `test_mask_batch_shapes_forward`,
`test_float_mask_batch_shape_differentiable`, `test_batch_mask_value_correctness`,
and `test_batch_mask_gradient_correctness` are parametrized over
`MASK_BATCH_SHAPES × EVENT_SHAPES`. The gradient test verifies
`grad[*b, k] = sigmoid'(0) * sum_e(act[k,e] - obs[*b,e])` across the full grid.

chirho/counterfactual/handlers/counterfactual.py
rfl-urbaniak: Holding off on unconditional sigmoid. The intended use case is composing with upstream categorical sampling: log probs → soft top-k → hard 0/1 mask via STE → `BatchedAction`. The 0s and 1s need to mean "factual" and "fully counterfactual" for that chain to work — `sigmoid(0) = 0.5` and `sigmoid(1) ≈ 0.73` break the intended semantics.

Dropping `validate_args` as suggested.

jfeser (approved):

